PR #10, opened 2014-05-15
Cassandra fails to install, wrong file name

amontalban:
The right file name is %%CQL%%%%DATADIR%%/lib/cql-internal-only-1.4.1.zip below is the build error:

===>  Installing for cassandra-1.2.15
===>  Checking if databases/cassandra already installed
===>   Registering installation for cassandra-1.2.15
pkg-static: lstat(/var/tmp/usr/ports/databases/cassandra/work/stage/usr/local/share/cassandra/lib/cql-internal-only-1.4.0.zip): No such file or directory
*** Error code 74

Stop.
make: stopped in /usr/ports/databases/cassandra

### DIFF
--- a/README
+++ b/README
@@ -15,7 +15,7 @@ These will explain how to use ports and packages.
 
 If you would like to search for a port, you can do so easily by
 saying (in /usr/ports):
-
+asdasdasdasd
 
 	make search name="<name>"
 	or:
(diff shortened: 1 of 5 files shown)


mat813:
databases/cassandra is not there any more.

PR #11, opened 2014-09-14
Delete patch-src-jrd_isc_sync.cpp

mariuz:
Patch is not needed in Firebird 2.5.3

Default number of SEMMNS  is 340 in FreeBSD 9/10

http://tracker.firebirdsql.org/browse/CORE-3400
http://sourceforge.net/p/firebird/mailman/message/27734396/
http://fxr.watson.org/fxr/source/kern/sysv_sem.c?v=FREEBSD10#L142


uqs:
Please file this as a PR, see https://www.freebsd.org/support/bugreports.html

PR #12, opened 2014-09-14
py-tipper, get python stack tracebacks via SIGUSR1

splbio:


mat813:
This was committed long ago.

PR #19, opened 2014-11-06
Add uwsgi_emperor support

Kronuz:
Emperor mode was not supported, this pull request adds support to it.
Use by setting uwsgi_emperor="/directory/to/uwsgi/vassals"

### DIFF
--- a/www/uwsgi/files/uwsgi.in
+++ b/www/uwsgi/files/uwsgi.in
@@ -24,6 +24,8 @@
 #				Default is "-M -L".
 # uwsgi_procname (str):		Define to "uWSGI" if you start uwsgi with
 #				--auto-procname option.
+# uwsgi_emperor (str):		Setup uwsgi to run as emperor in the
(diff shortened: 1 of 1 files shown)


mat813:
Openned a PR https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=210170

Dafvid:
Still not committed =/

Dafvid:
Trying to make this work but I have issues.
- rc is spawning two processes and I can't seem to kill both. why 2?
- is there a better way than rm -f to stop the process ('uwsgi stop <pid>')?
- how about just open up for passing a path to config and set everything up in there instead?

rakuco:
It'd be great if you could post comments to Bugzilla instead. The port maintainer is not watching this pull request (which was closed last year) and any further discussions here are likely to get ignored by the relevant parties. FWIW I'm also CC'ed in Bugzilla and can help you land this once the issues are solved.

PR #23, opened 2014-12-31
virtualbox-ose-lite minimal deps for packaging

splbio:
This package has minimal deps, only udp server and
vncserver are required.

This works really nicely for someone who just wants
to run headless vms or the VBoxManage tools without
bloat.

### DIFF
--- a/emulators/Makefile
+++ b/emulators/Makefile
@@ -148,6 +148,7 @@
     SUBDIR += virtualbox-ose
     SUBDIR += virtualbox-ose-additions
     SUBDIR += virtualbox-ose-kmod
+    SUBDIR += virtualbox-ose-lite
     SUBDIR += visualboyadvance-m
     SUBDIR += vmips
     SUBDIR += vmsbackup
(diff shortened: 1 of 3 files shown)


splbio:
Why is this closed?  The port was asked for by the re@ team.

mat813:
Because we don't do pull requests, if you want to submit something, open a PR or a review on phabricator.

splbio:
What do you mean by "we"?  I'm a freebsd developer and a polling at the recent meetbsd2015 showed that fully 2/3 of the room uses git.  A lot of "we" use git and github.  Thanks!

mat813:
First, let's say that it's 18 months old, nobody seemed to care enough to commit it, so closing it is right.

Then, we being the ports people. I also use git for all my ports work, but there's no point in doing pull requests, we won't "merge" here, it would break the importer.

To use a pull request, one has to fiddle around with git fetch to get the pull request branch, then one has to cherry pick the commit(s) if there's more than one commit, one has to squash them. Then, one will most likely have to clean things up, and squash again, then, one will be able to git svn dcommit.

It would have been faster to open a PR with the patch from here, feel free to do so.

splbio:
Is these still no tool to port these pull requests over to PRs or directly submit them? :(

PR #26, opened 2015-02-04
Add Atom and Brackets

finid:
Couldn't find Atom (https://atom.io/) and Brackets (http://brackets.io/) in the ports tree on a new installation of PC-BSD 10.1.1 or even on he Web interface of the ports tree. Those are very good Markdwon editors and general-purpose IDEs. Would be great to have them in ports.


mat813:
Closing this, it has way too much commits to make sense.

PR #27, opened 2015-03-11
Upgrade to Boost 1.57

windoze:
Upgrade to Boost 1.57.
All patch files in boost-libs are removed as they're all merged into Boost source code or not applicable.


bapt:
moved to github and breaks lots of existing ports see: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=199601

PR #28, opened 2015-04-21
Upgrade Boost to 1.58

windoze:
Upgrade Boost to 1.58


bapt:
See: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=199601

PR #29, opened 2015-05-15
Added -m to display in MB/s or Mb/s (when -b is used)

jshupe:


mmoll:
Could you try to get this into the upstream project and a new version to get released?
If this turns out to be not possible, please ping me here.

mat813:
This belongs upstream -> http://gael.roualland.free.fr/ifstat/

PR #30, opened 2015-05-19
Elixir always expects UTF-8, make it default while compiling to avoid…

dinapappor:
… warning messages.

Elixir needs and expects UTF-8 and emits ```warning: the VM is running with native name encoding of latin1 which may cause Elixir to malfunction as it expects utf8. Please ensure your locale is set to UTF-8 (which can be verified by running "locale" in your shell)``` if not compiled with UTF-8.

This PR fixes that.


mat813:
In the Makefile there is:

```
MAKE_ENV=       LANG=en_US.UTF-8
```

Which I believe fixes the issue.

PR #33, opened 2015-08-13
Update bsd.php.mk

Tequila721:
I've add redis into _USE_PHP_ALL  & redis_DEPENDS in bsd.php.mk to solved when I put redis to php makefile and install will got error : unknown extension redis for php


mmoll:
As CONTRIBUTING.md states "FreeBSD does not currently accept pull requests. This is a read-only mirror from SVN.". Please see http://www.freebsd.org/doc/en_US.ISO8859-1/books/porters-handbook/porting-submitting.html

mat813:
I opened a PR https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=210171

PR #35, opened 2015-09-18
py-pytest-runner: remove the version 2 forcing

nerzhul:
This fix the py34-pytest-runner build, and by side the py34-irc port build.


opsec:
There are a few other ports that depend on this:
devel/py-jaraco.timing
security/py-keyring
net/py-wolframalpha
Can they cope with that change as well ?

nerzhul:
@opsec for devel/py-jaraco.timing & security/py-keyring no problem they are not python 2 or 3 dependant.
For net/py-wolframalpha the port is forced to v2. It will work on python 2.7 version, but not in 3.4, it's same problem as this port, it could not be built. Then there is no problem affecting py-pytest-runner here

opsec:
Change was committed in r397529, see also

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=203248

PR #36, opened 2015-10-06
x11/libxkbcommon: Set correct x-locale path via CONFIGURE_ARGS.

ivadasz:
This enables libxkbcommon to find the locale-specific key
composition rules installed by x11/libX11.


ivadasz:
sorry for the noise, didn't notice this was read-only ...

PR #38, opened 2015-11-15
Add start_x.elf and fixup_x.dat files.

serge-v:
This fix adds missing firmware start_x.elf and fixup_x.dat files to boot partition for Raspberry PI (versions A and B). This is required for camera and additional codecs support.


mat813:
This looks a lot like https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=210041

PR #41, opened 2016-01-17
Updated FreeBSD port to build violetland 0.5

ooxi:
[Violetland](https://github.com/ooxi/violetland) was recently released in version 0.5. Since @AMDmi3 is the maintainer of this port it would be great if he could merge the changes.

I originally [created a patch](https://github.com/ooxi/violetland-freebsd-port/commit/156d859b17fc1374f51f59ad295ea31055ce13ae.patch) but then saw that the FreeBSD ports collection is available on GitHub.


AMDmi3:
We do not currently accept ports changes via pull requests, you should've sent a patch through our [bugzilla](https://bugs.freebsd.org/bugzilla/). Actually, I was already working on an update, and it requires some more changes than in your patch. However before I wanted to submit some chenges to the repo, because gettext detection and usage is broken. For instance, heavy external modules seem to be used just for gettext detection, while there's FindGettext bundled with CMake.

ooxi:
Thanks for the input! I do not remember why I didn't use the bundled `FindGettext` from CMake but I will evaluate again. My goal is to enable building on FreeBSD without much modification.

mat813:
The maintainer is working on this, so close.

PR #42, opened 2016-01-27
Update nzbget.in

dburkland:
Fix broken status command which was returning "eval: -L: not found" on my stock FreeBSD 10.2 system. Simply moved the "command" variable definition before the *_cmd variables.


dburkland:
Pushing my fix for the nzbget init script which currently has a broken status command.

mat813:
This was committed.

PR #43, opened 2016-02-03
security/py-letsencrypt: Fix pkg-message

liquid:
Only a word duplication fix but this is my first contribution to the ports tree :)


mat813:
Committing this.

PR #45, opened 2016-02-24
add 3rd party upstream_health_check module to nginx

ctriegg:
I'm not shure how to ask for adding those changes to the ports tree as this git repository is just a mirror of the FreeBSD-ports SVN repo. But I found this page https://wiki.freebsd.org/GitWorkflow/GitSvn. Is this the appropiate way? I found nothing about contribution in the porter's handbook.


mat813:
The usual way of doing things is to submit a bug on our bugzilla, see https://www.freebsd.org/doc/en/books/porters-handbook/port-upgrading.html
I have opened a bug report for you:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=210172

PR #49, opened 2016-05-26
Add filebeat and packetbeat from Elastic

rcalixte:
@girgen has a working deployment for filebeat and packetbeat (Latest version is 1.2.3, but 1.1.2 is still compatible)
https://github.com/elastic/beats

Would be nice to have this upstream (@opsec @mat813)


mat813:
`lang/go/files/bsd.go.mk` doesn't exist any more, it's called USES=go now, could you update the patch ?

rcalixte:
I've opened a PR to @girgen with the patch. I may open a PR from the fork if it waits a while.

girgen:
Hi,

I'm not certain that separate ports is ideal.

sysutils/filebeat
sysutils/packetbeat
sysutils/topbeat

or just

sysutils/beats

thoughts?

Palle

rcalixte:
There are other community beats here (https://www.elastic.co/guide/en/beats/libbeat/master/community-beats.html) so I would think that separate ports is the correct structure as we're not going to roll them all up.

girgen:
Great!

Did you try the ports? The reason I have not committed is that I haven't got around to using them yet, so I really only know that they build properly... :)

But If you say they work, I can commit them to the (real svn) tree.

Palle

mat813:
Like I said, in the current form, they won't.

opsec:
I modified the submitted patches and I'm testbuilding right now.

opsec:
Is there a PR in bugs.freebsd.org somewhere for these ports ?

rcalixte:
I've opened https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=209776 for these ports.

rcalixte:
Both filebeat and packetbeat are in the ports tree now. Closing. Thanks all!

PR #50, opened 2016-06-30
SUCCESSFUL CROSS-BUILD with some minor bugs

epipenau:
is hard coded to cater for freebsd-wifi-build world because currently xdev is broken.

every port i've tried to build with this so far works.  but there are failures in other areas, test it out to see.


mat813:
There is a review for this, so close here.

koobs:
[FreeBSD Review D7049: Improve cross building](https://reviews.freebsd.org/D7049)

PR #52, opened 2016-08-22
Build with native PKCS#11 Interface 

hb9cwp:
BIND-9.10+ provides a native PKCS#11 client [1] as a simpler and more recent alternative to interface with PKCS#11 providers, such as SoftHSMv2, compared to using High Security Modules (HSM) via OpenSSL.

The blog [2] illustrates how to use BIND's ```pkcs11-*``` and ```dnssec-* tools```  in native PKCS11 mode to generate DNSSEC Key and Zone Signing Keys in a SoftHSMv2, and to sign zones' resource records. 

This PR includes patches which I have hand-picked from Fedora 23 port's ```bind-9.10.4-1.P2.fc23.src.rpm``` patch file ```.../rpmbuild/SOURCES/bind-9.10-dist-native-pkcs11.patch```.

Note that I removed DLZ while closely following the "native PKCS11" patch of the Fedora port, although presently unsure if this necessary. 

Maybe it is cleaner, or even necessary, to have separate port like "bind910-pkcs11..." in Linux, or a "flavour" thereof like OpenBSD ports system provides?

This is my first PR for FreeBSD, please be kind :-)   Thanks.

[1] Native PKCS#11 mode
 https://ftp.isc.org/isc/bind9/cur/9.10/doc/arm/Bv9ARM.ch04.html#id-1.5.12.7

[2] DNSSEC with BIND 9.10 and native PKCS#11 support (BIND and SoftHSM)
 http://arunnsblog.com/2016/01/18/dnssec-with-bind-9-10-and-native-pkcs11/


hb9cwp:
see discussion at https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=212069

mat813:
Close this here.

PR #53, opened 2016-09-27
Fix typo which remove disk plugin everywhere

vr:


jbeich:
@Jehops fixed it by 77373edbf87b. CC @swills to close.

PR #54, opened 2016-10-04
Fix an issue in vagrant when path to key contains '%'

trombik:
this pacth fixes https://github.com/mitchellh/vagrant/issues/7863

https://github.com/mitchellh/vagrant/pull/7862


trombik:
sorry wrong tree

PR #57, opened 2016-11-06
Update MAINTAINER field

harmw:
Bert will be taking over 👍

### DIFF
--- a/net/cloud-init/Makefile
+++ b/net/cloud-init/Makefile
@@ -6,7 +6,7 @@ CATEGORIES=	net python
 MASTER_SITES=	http://launchpad.net/${PORTNAME}/trunk/${PORTVERSION}/+download/
 PKGNAMEPREFIX=	${PYTHON_PKGNAMEPREFIX}
 
-MAINTAINER=	harm@weites.com
+MAINTAINER=	bertjw@regeer.org
 COMMENT=	Init scripts for use on cloud images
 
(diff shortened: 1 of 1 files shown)


jbeich:
I can't verify you're the current maintainer and not a doppelganger. Ports use mail addresses to identify maintainers, not real names or accounts outside of `.freebsd.org` domains. Our Bugzilla installation automates such workflow but on Github... not sure how to do it without exposing mail to spammers.

net/cloud-init/Makefile
```diff
@@ -6,7 +6,7 @@ CATEGORIES=	net python
 MASTER_SITES=	http://launchpad.net/${PORTNAME}/trunk/${PORTVERSION}/+download/
 PKGNAMEPREFIX=	${PYTHON_PKGNAMEPREFIX}
 
-MAINTAINER=	harm@weites.com
+MAINTAINER=	bertjw@regeer.org
```
jbeich: This address appears to be different from the one used in [bug 212453](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=212453). Is it still the same person?

jbeich:
Also, do you have Bert's consent?

harmw:
Yes I have, he actually offered to take over. I pinged him a day ago, referring to this PR. As you can see, my GH username doesn't differ much from the e-mail address being replaced 😅

PR #58, opened 2016-11-15
Fix python35 openssl devel

melvyn-sopacua:
Make it build with security/openssl-devel by applying patches from Python issue 26470.


uqs:
Please file this as a PR, see https://www.freebsd.org/support/bugreports.html

PR #59, opened 2016-11-26
Fixed sysrc command in instructions

DadAtH-me:
Command as listed in documentation:

    root@zm:/usr/local/share/mysql # sysrc mysql_server_enable="YES"
    mysql_server_enable:  -> YES
    root@zm:/usr/local/share/mysql # service mysql-server start
    Cannot 'start' mysql. Set mysql_enable to YES in /etc/rc.conf or use 'onestart' instead of 'start'.

Change:

    root@zm:/usr/local/share/mysql # sysrc mysql_enable="YES"
    mysql_enable:  -> YES
    root@zm:/usr/local/share/mysql # service mysql-server start
    Starting mysql.


novel:
Hi, this is just a read-only mirror.

Please use Bugzilla for submitting problem reports and patches: https://www.freebsd.org/support/bugreports.html

Thanks

PR #64, opened 2017-07-22
net-p2p/transmission: Version bump.

lebarondemerde:
Updating the meta port to reflect the actual versions of the net-p2p/transmission-* ports.


lebarondemerde:
Sorry,
I forgot to bump PORTREVISION, and so I preferred to create a complete new one instead of two minors.

PR #66, opened 2017-08-17
sysutils/goss: Add rc script, sample configuration (221579)

englishm-llnw:
Adds an rc script to run `goss serve` as a health endpoint and sample configuration files to support the service.

Reference only.

Actual patch submitted as [Bug 221579](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=221579)

### DIFF
--- a/sysutils/goss/Makefile
+++ b/sysutils/goss/Makefile
@@ -3,6 +3,7 @@
 PORTNAME=		goss
 PORTVERSION=		0.3.4
 DISTVERSIONPREFIX=	v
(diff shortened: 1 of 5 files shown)


englishm-llnw:
Committed by @swills as [r448105](https://svnweb.freebsd.org/ports?view=revision&revision=448105)
In this Git mirror as: https://github.com/freebsd/freebsd-ports/commit/b4e830ac6aeb57fb35b836a971f34bbba005bde8

PR #74, opened 2017-11-20
Update pkg-messages in

hachiman8:
Fix URL


swills:
We don't accept pull requests here, please open an issue at:

https://bugs.freebsd.org/bugzilla/

and attach the patch file to the issue.

opsec:
Thanks for the patch, fixed with commit r455694.

PR #81, opened 2018-03-01
net/google-cloud-sdk: Update Makefile, Fix Bug 225255

ClearPathDigital:
Update version to 191.0.0,
Add run dependency databases/py27-sqlite3.
Fixes: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=225255


ClearPathDigital:
Disregard, this patch needs more work...

PR #98, opened 2019-04-29
UBoot slave file for Odroid XU3 and NanoPi NEO2

nekoprog:


evadot:
Hi,

I've just commited the nanopi-neo2 u-boot (https://svnweb.freebsd.org/ports?view=revision&revision=500658) which was submitted by phabricator.
I don't see a point of commiting the Odroid Xu3 since we don't run (or barely run on this board).
Thanks.

nekoprog:
If we have odroid support, it would be great, but no problem. Can we have nanopi-r1 too?

evadot:
Please submit a review on review.freebsd.org for that and assign it to me.
Thanks

nekoprog:
@evadot Can you add for Odroid XU4? I uses same defconfig and u-boot salve port as Odroid XU3, it boast great specs for 32bit arm device.

Product page: https://www.hardkernel.com/shop/odroid-xu4-special-price/

PR #99, opened 2019-07-06
semun errors

metacreo:
semun errors


metalefty:
FreeBSD does not currently accept pull requests. This is a read-only mirror from SVN.

See this document and submit your patch to FreeBSD bugzilla.
* https://www.freebsd.org/doc/en_US.ISO8859-1/books/porters-handbook/porting-submitting.html

PR #109, opened 2021-04-15
S3 files should ONLY appear in -server not -client

darkain:
This line is causing -client and -server to conflict with each other by both attempting to install the same file to the same location. S3 is a -server storage engine item only, and has no relation to the -client package.

Referenced in:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=251622
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=254969


kev009:
0d06019248512 handled this

PR #110, opened 2021-04-22
Allow Overlay UIDs and GUIDs

Dark-Schnitzel:
Allow the overlay of UIDs and GUIDs.
https://lists.freebsd.org/pipermail/freebsd-ports/2021-January/120134.html


mat813:
Mmmm, OVERLAYS is plural, there can be more than one, so, there should be loops somewhere in there.

jaceppo (approved):

heroku-miraheze:
https://docs.freebsd.org/en/books/porters-handbook/ is the page where to contribute.

heroku-miraheze:
This repo is read-only, and PRs is not allowed.

PR #111, opened 2021-05-11
Modularize OpenLDAP server

zsalab:
The default configuration of OpenLDAP server has no overlays and modularized just from a backend perspective. I think would be more user-friendly if the backends and the overlays should be modularized and by default added to the package.
With this change, the default build will contain most of the backend and overlay modules as dynamic modules.
This may cause a compatibility issue as by default only the SYNPROV overlay was compiled into the server (as static) so anybody already using OpenLDAP from the repo may have to enable the module after install.
Would be nice to get suggestions on compatibility improvements.

P.S.: I just started to use BSD, so maybe I am complete on the wrong path with this change but I am happy to learn


delphij:
Committed with some minor changes (mainly: slightly different and expanded default, and changed PROXYCACHE to PPCACHE to match filename being installed, which will make maintenance of the default options easier).

PR #115, opened 2021-10-10
Update package descriptions and messages for some VirtualBox ports

grahamperrin:
Package messages are more focused on: 

* what's recommended and/or required
* how to avoid system crashes of guests with non-legacy additions.

(Whilst it's possible for the same crash to occur with legacy additions, incidents may be so rare that it's not worth mentioning in the package message. The bug can be found through a click on the bug icon at FreshPorts.)

Package messages are more specific. 

Package messages no longer refer to <https://wiki.freebsd.org/VirtualBox>. There's not much of substance in the wiki – users might as easily find the page through, for example, attention to the FreeBSD Handbook. 

Draft; comments and corrections are invited. Thank you.

### DIFF
--- a/emulators/virtualbox-ose-additions-legacy/pkg-descr
+++ b/emulators/virtualbox-ose-additions-legacy/pkg-descr
@@ -1,9 +1,17 @@
-These additions are for installation inside a FreeBSD guest.
+For any serious and interactive use, the Oracleⓡ VM VirtualBoxⓡ 
+Guest Additions will make your life much easier by providing closer 
+integration between host and guest and improving the interactive 
+performance of guest systems. 
 
-VirtualBox is a family of powerful x86 virtualization products for
-enterprise as well as home use. Not only is VirtualBox an extremely
-feature rich, high performance product for enterprise customers, it
-is also the only professional solution that is freely available as
-Open Source Software under the terms of the GNU General Public License.
+Non-legacy additions i.e. emulators/virtualbox-ose-additions are 
+recommended for most use cases. 
 
-WWW: https://www.virtualbox.org/
+Where legacy (5.2.x) is a requirement, this FreeBSD-provided package 
+can be installed in the FreeBSD guest machine. 
+
+<https://www.virtualbox.org/wiki/Changelog-5.2> emphasises that 
+5.2.x is no longer supported by Oracle.
+
+Addition to host machines is not required. 
+
+WWW: https://www.virtualbox.org/manual/UserManual.html#guestadditions
--- a/emulators/virtualbox-ose-additions-legacy/pkg-descr
+++ b/emulators/virtualbox-ose-additions-legacy/pkg-descr
@@ -1,7 +1,5 @@
-For any serious and interactive use, the Oracleⓡ VM VirtualBoxⓡ 
-Guest Additions will make your life much easier by providing closer 
-integration between host and guest and improving the interactive 
-performance of guest systems. 
+VirtualBox Guest Additions provide closer integration between host and 
+guest, and improve the performance of guest systems. 
 
 Non-legacy additions i.e. emulators/virtualbox-ose-additions are 
 recommended for most use cases. 
@@ -14,4 +12,4 @@ can be installed in the FreeBSD guest machine.
 
 Addition to host machines is not required. 
 
-WWW: https://www.virtualbox.org/manual/UserManual.html#guestadditions
+WWW: https://www.virtualbox.org/
--- a/emulators/virtualbox-ose-additions-legacy/pkg-message
+++ b/emulators/virtualbox-ose-additions-legacy/pkg-message
@@ -1,7 +1,7 @@
 [
 { type: install
   message: <<EOM
-Oracleⓡ VirtualBoxⓡ Guest Additions are installed.
+VirtualBox Guest Additions are installed.
 
 To enable and start the required services: 
 
--- a/emulators/virtualbox-ose-additions-legacy/pkg-message
+++ b/emulators/virtualbox-ose-additions-legacy/pkg-message
@@ -1,23 +1,24 @@
 [
 { type: install
   message: <<EOM
-VirtualBox Guest Additions were installed.
+Oracleⓡ VirtualBoxⓡ Guest Additions are installed.
 
+To enable and start the required services: 
 
-You need to enable the vboxguest startscript to load the kernel module and
-vboxservice to use host time synchronization.
+# sysrc vboxguest_enable="YES" && service vboxguest start
+# sysrc vboxservice_enable="YES" && service vboxservice start
 
-vboxguest_enable="YES"
-vboxservice_enable="YES"
+The vboxguest service will load a required kernel module. 
 
-You also have to add all X11 users that want to use any of the additional
-features (clipboard sharing, window scaling) to the wheel group.
+For features such as window scaling and clipboard sharing, membership of 
+the wheel group is required. With username jerry as an example: 
 
-% pw groupmod wheel -m jerry
+# pw groupmod wheel -m jerry
 
-Reboot the machine to load the needed kernel modules.
-
-For detailed informations please visit http://wiki.freebsd.org/VirtualBox
+The settings dialogue for FreeBSD guests encourages use of the VMSVGA 
+graphics controller. Whilst this might suit installations of FreeBSD 
+without a desktop environment (a common use case), it may be inappropriate 
+where legacy Guest Additions are installed. 
 EOM
 }
 ]
--- a/emulators/virtualbox-ose-additions-legacy/pkg-message
+++ b/emulators/virtualbox-ose-additions-legacy/pkg-message
@@ -5,8 +5,8 @@ Oracleⓡ VirtualBoxⓡ Guest Additions are installed.
 
 To enable and start the required services: 
 
-# sysrc vboxguest_enable="YES" && service vboxguest start
-# sysrc vboxservice_enable="YES" && service vboxservice start
+# sysrc vboxguest_enable="YES" ; sync ; service vboxguest start
+# sysrc vboxservice_enable="YES" ; service vboxservice start
 
 The vboxguest service will load a required kernel module. 
 
--- a/emulators/virtualbox-ose-additions-legacy/pkg-message
+++ b/emulators/virtualbox-ose-additions-legacy/pkg-message
@@ -3,12 +3,12 @@
   message: <<EOM
 VirtualBox Guest Additions are installed.
 
-To enable and start the required services: 
+To enable the required services: 
 
-# sysrc vboxguest_enable="YES" ; sync ; service vboxguest start
-# sysrc vboxservice_enable="YES" ; service vboxservice start
+# sysrc vboxguest_enable="YES"
+# sysrc vboxservice_enable="YES"
 
-The vboxguest service will load a required kernel module. 
+To start the services, restart the system. 
 
 For features such as window scaling and clipboard sharing, membership of 
 the wheel group is required. With username jerry as an example: 
--- a/emulators/virtualbox-ose-additions/pkg-message
+++ b/emulators/virtualbox-ose-additions/pkg-message
@@ -1,7 +1,7 @@
 [
 { type: install
   message: <<EOM
-Oracleⓡ VirtualBoxⓡ Guest Additions are installed.
+VirtualBox Guest Additions are installed.
 
 To enable and start the required services: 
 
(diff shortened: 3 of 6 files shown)


emulators/virtualbox-ose-additions-legacy/pkg-descr
```diff
@@ -1,9 +1,17 @@
-These additions are for installation inside a FreeBSD guest.
+For any serious and interactive use, the Oracleⓡ VM VirtualBoxⓡ
```
madpilot78, line 1: While I can't cite a written rule for this, better avoid special characters and stick to plain ascii as much as possible.

I'm actually not sure (r) is correct here, I think (tm) for trademark is what you'd want, but I don't see the need for that information here.

emulators/virtualbox-ose-additions-legacy/pkg-descr
```diff
@@ -1,9 +1,17 @@
-These additions are for installation inside a FreeBSD guest.
+For any serious and interactive use, the Oracleⓡ VM VirtualBoxⓡ 
+Guest Additions will make your life much easier by providing closer
```
madpilot78, line 2: I also would try a less opinionated expression. Just inform the user that the port installs additions, don't assume what "serious" use is. I mean these parts sound a little too much like you're trying to "sell" the port. That is not the point of pkg-descr.

Something like "Installs tools to make interactive use of VirtualBox easier to use by... (etc)

emulators/virtualbox-ose-additions-legacy/pkg-descr
```diff
@@ -1,9 +1,17 @@
...
-WWW: https://www.virtualbox.org/
+Where legacy (5.2.x) is a requirement, this FreeBSD-provided package 
+can be installed in the FreeBSD guest machine. 
+
+<https://www.virtualbox.org/wiki/Changelog-5.2> emphasises that 
+5.2.x is no longer supported by Oracle.
+
+Addition to host machines is not required. 
+
+WWW: https://www.virtualbox.org/manual/UserManual.html#guestadditions
```
madpilot78, line 17: Please stick to the home page of the product with the WWW link.

emulators/virtualbox-ose-additions-legacy/pkg-message
```diff
@@ -1,23 +1,24 @@
 [
 { type: install
   message: <<EOM
-VirtualBox Guest Additions were installed.
+Oracleⓡ VirtualBoxⓡ Guest Additions are installed.
```
madpilot78, line 4: Again try avoiding non ascii characters for maximum compatibility. Non need to put "Oracle" here, the software name is quite enough.

emulators/virtualbox-ose-additions-legacy/pkg-message
```diff
@@ -1,23 +1,24 @@
...
+# sysrc vboxservice_enable="YES" && service vboxservice start
 
-vboxguest_enable="YES"
-vboxservice_enable="YES"
+The vboxguest service will load a required kernel module. 
 
-You also have to add all X11 users that want to use any of the additional
-features (clipboard sharing, window scaling) to the wheel group.
+For features such as window scaling and clipboard sharing, membership of 
+the wheel group is required. With username jerry as an example:
```
madpilot78, line 14: AFAIK it should be "membership to the wheel group", but I'm not a native English speaker so my opinion on this is not definitive.

emulators/virtualbox-ose-additions-legacy/pkg-message
```diff
@@ -1,23 +1,24 @@
...
-% pw groupmod wheel -m jerry
+# pw groupmod wheel -m jerry
 
-Reboot the machine to load the needed kernel modules.
-
-For detailed informations please visit http://wiki.freebsd.org/VirtualBox
+The settings dialogue for FreeBSD guests encourages use of the VMSVGA 
+graphics controller. Whilst this might suit installations of FreeBSD 
+without a desktop environment (a common use case), it may be inappropriate 
+where legacy Guest Additions are installed.
```
madpilot78, line 21: This is about the legacy additions, I have not experimented but I think in this case the best choice is the old VBoxVGA driver.

grahamperrin: Thanks. 

Shall we write VBoxVGA into the message? Then if someone who does experiment finds fault, the message can be reviewed.

emulators/virtualbox-ose-additions-legacy/pkg-descr
grahamperrin: Thanks. 

I'm ambivalent about use of non-ASCII characters in the context of files such as `pkg-descr` and `pkg-message`. 

_Oracle® VM VirtualBox®_ appears at the head of Oracle's [User Manual](https://www.virtualbox.org/manual/UserManual.html).

grahamperrin: ### Generally

I aim to treat Oracle-related texts in much the same way as I would _any_ other text that includes a trademark. (It's not my intention to give special treatment to Oracle.)

<https://www.oracle.com/legal/trademarks.html> "… Use the ® symbol with the most prominent appearance of the "Oracle" mark on products, packaging, manuals, advertisements, promotional materials and Web pages (for example, in the headline of an advertisement), and the first use of the mark in text or body copy. …", and the package descriptions _are_ represented within web pages, **however** I understand that some readers might view this as over-the-top.

emulators/virtualbox-ose-additions-legacy/pkg-descr
grahamperrin: Here, I'm torn. 

Whilst Guest Additions can not be used in isolation from the broader VirtualBox product, I'd like the 'main' link to be as close as possible to a home page for Guest Additions. 

In the absence of a home page for Guest Additions, <https://www.virtualbox.org/manual/UserManual.html#guestadditions> (within the User Manual) seems appropriate. 

If this wish is contrary to FreeBSD tradition for use of `WWW` in package descriptions: sorry! 

-----

It's probably fair to say that I'm being more _realistic_ than _traditional_. 

For a newcomer to VirtualBox, the home page for VirtualBox does not offer a quick 'stepping stone' to Guest Additions … and so on.

emulators/virtualbox-ose-additions-legacy/pkg-message
grahamperrin: Reasoning at https://github.com/freebsd/freebsd-ports/pull/115/files#r730346438 above. 

I'm quite ambivalent (neutral). Ultimately: 

> … I understand that some readers might view this as over-the-top.

emulators/virtualbox-ose-additions-legacy/pkg-descr
madpilot78: You're missing the main point I made, while there is not a strict rule about this, pkg-descr files can be read from any terminal (even old ones) which can be configured with any kind of locale and convention. SO there is no warranty about the ability to understand and properly display UTF-8 (or any other encoding) characters.

These files should be kept strictly pure ASCII for maximum compatibility.

Due to this if you really think the "registered" symbol is necessary, please render it as (r) or (R), which is an accepted convention.

Regarding the secondary point, my personal opinion is that such symbols here are totally unwarranted for, for reasons that are not technical and are beyond this discussion. Anyway I can get past this opinion of mine and include them when committing as long as they do not interfere with compatibility when displaying the file.

madpilot78: I'd add that it is normal for oracle to want a commercial rendition of their company name always attached to the name of the product and clearly specify it's a registered trademark.

We, on the other hand, are only using that name to reference the product.

emulators/virtualbox-ose-additions-legacy/pkg-descr
madpilot78: The official user manual is something produced by oracle who has all interest in promoting their product, so it's ok for it to sound like marketing material. This does not mean we must report that varbatim.

A suggestion is:

The Oracle (r) VirtualBox (r) additions allow for closer integration between host and guest operating systems, allowing for a better and easier user experience, also helping with performance

(BTW the original description contains a mistake IMHO, omitting the "operating system" part about the integration, but that may be intentional, again for marketing purposes)

emulators/virtualbox-ose-additions-legacy/pkg-message
madpilot78: As above, it's about compatibility, not about adversity(which I do have, but I will not push) towards such symbols, please use the ASCII rendition of such symbols.

emulators/virtualbox-ose-additions-legacy/pkg-message
madpilot78: I accept this, I was only expressing a cursory doubt about this.

emulators/virtualbox-ose-additions-legacy/pkg-message
grahamperrin: Thanks for guidance. Much appreciated.

emulators/virtualbox-ose-additions-legacy/pkg-message
```diff
@@ -1,23 +1,24 @@
...
 { type: install
   message: <<EOM
-VirtualBox Guest Additions were installed.
+VirtualBox Guest Additions are installed.
 
+To enable and start the required services: 
 
-You need to enable the vboxguest startscript to load the kernel module and
-vboxservice to use host time synchronization.
+# sysrc vboxguest_enable="YES" ; sync ; service vboxguest start
```
madpilot78, line 8: Is really this sync command required on modern systems? I think it can be confusing and in the rare case a system crashes in the middle of the commands the user can anyway perform them gain.

This operation is not atomic anyway, but sysrc is idempotent if ran multiple times.

grahamperrin: > why would it be necessary in the first command and not the second?

"The vboxguest service will load a required kernel module." 

If there's no kernel panic at load time, then there's no need to sync after what follows. 

> How can you be sure the crash was related to you launching sysrc and not to something else that was going on at the time?

<https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=254412#c7>

emulators/virtualbox-ose-additions-legacy/pkg-message
```diff
@@ -3,12 +3,12 @@
...
-To enable and start the required services: 
+To enable the required services: 
 
-# sysrc vboxguest_enable="YES" ; sync ; service vboxguest start
-# sysrc vboxservice_enable="YES" ; service vboxservice start
+# sysrc vboxguest_enable="YES"
+# sysrc vboxservice_enable="YES"
 
-The vboxguest service will load a required kernel module. 
+To start the services, restart the system.
```
madpilot78, line 11: You can also state that users can also start the services manually, it's really up to the user.

emulators/virtualbox-ose-additions/pkg-message
```diff
@@ -5,12 +5,12 @@ VirtualBox Guest Additions are installed.
...
-# sysrc vboxservice_enable="YES" ; service vboxservice start
+# sysrc vboxguest_enable="YES"
+# sysrc vboxservice_enable="YES"
 
-The vboxguest service will load a required kernel module. If a 
-kernel panic (system crash) occurs, this may be avoided by 
-limiting the FreeBSD guest to one CPU. 
+To start the services, restart the system. If a kernel panic (system crash) 
+occurs during startup, this may be avoided by limiting the FreeBSD guest to 
+one CPU.
```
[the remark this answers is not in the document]
grahamperrin: https://wiki.freebsd.org/Graphics#Virtual_Machines for VMWare mentions "A race condition … Observable only if the guest is configured with more than one CPU.". 

Shall we move this comparable observation for VirtualBox – about a kernel panic, observable only if the guest is configured with more than one CPU – to <https://wiki.freebsd.org/VirtualBox>?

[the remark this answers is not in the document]
grahamperrin: > … a sure fix, which is too much. …

To me, "can" is surer than "may" – that's _may_ as in _maybe_. 

How about: 

"In some situations, a panic will occur when the kernel module loads. Having no more than one virtual CPU might mitigate the issue."

madpilot78:
Changes committed. This pull request should be closed. Github does not allow me to do that.

grahamperrin:
Thanks! 

Re: <https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=254412#c17> apologies for the trailing white space. 

(It's a wordprocessing habit from years gone by, mentally ingrained. I'll try to remember to avoid/undo the practice for future code.)

PR #116, opened 2021-11-03
Update just journal client to 2.0.1

laffer1:


metalefty:
This GitHub repo is a read-only mirror so we don't accept pull requests. File it to [Bugzilla](https://bugs.freebsd.org).

PR #119, opened 2021-12-23
Updated net/p5-Net-Amazon-S3 from 0.98 to 0.99.

oodler577:
Regular version bump, fixed some minor upstream issues;
CPAN's tests show FreeBSD versions all in green for this
version of the module.

Please advise if this is not the proper way to update a port such as this.


oodler577:
Closing this PR to go through bug/patch submission webpage

PR #121, opened 2022-01-13
databases/influxdb: improve syslog logging title, remove trailing spaces in rc script

koitsu:
Currently, logging messages from influxd (which go through syslog) are logged as coming from `daemon` rather thant influxd itself.  Example:

> Jan 13 02:08:13 icarus daemon[43241]: ts=2022-01-13T10:08:13.768315Z lvl=info msg="InfluxDB starting" log_id=0Z0jOen0000 version=1.8.9 branch=unknown commit=unknown

This makes it impossible to filter said messages to a separate file via syslog.conf'd `!programname` feature.  And changing the facility to something like local7 is not sufficient either, as there are many programs which could use that facility as well.

This PR uses the `-T` flag of `daemon` to specify the process name, which alleviates this problem.  Example:

> Jan 13 02:09:10 icarus influxd[43292]: ts=2022-01-13T10:09:10.424810Z lvl=info msg="InfluxDB starting" log_id=0Z0jS760000 version=1.8.9 branch=unknown commit=unknown`

Thus, we can now successfully use a syslog.conf rule like the below to shove all influxdb messages into a separate file:

```
!influxd
*.*             /var/log/influxd.log
!*
```

Finally, this PR also removes a whole slew of trailing spaces from the rc script.

### DIFF
--- a/databases/influxdb/files/influxd.in
+++ b/databases/influxdb/files/influxd.in
@@ -43,25 +43,25 @@ influxd_pidfile="%%INFLUXD_PIDDIR%%${name}.pid"
 procname="%%PREFIX%%/bin/${name}"
 command=/usr/sbin/daemon
 start_precmd="influxd_precmd"
-start_cmd="influxd_startcmd_%%INFLUXD_LOGCMD%%"        
+start_cmd="influxd_startcmd_%%INFLUXD_LOGCMD%%"
 
 influxd_precmd()
 {
     install -d -o ${influxd_user} %%INFLUXD_PIDDIR%%
 }
 
-influxd_startcmd_daemon()                      
-{   
-    echo "Starting ${name}."                   
+influxd_startcmd_daemon()
+{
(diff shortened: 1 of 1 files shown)


driesmp:
Hello koitsu, change LGTM, I'll do some prep to get this in the tree.

driesmp:
> ```
> !influxd
> *.*             /var/log/influxd.log
> !*
> ```
Just an FYI, but the file above should already work currently, as the !influxd specifies the progname. *.* is here limited to daemon.*, as the facility is set to daemon.
The name tag we set here in this patch is only for the actual log entries. Which are much more clear after your patch to be coming from influxd :).

driesmp:
--author "Jeremy chadwick <e-mail goes here>"

Feel free to PM me your e-mail (driesm@freebsd.org), otherwise I'll commit without.

driesmp:
Committed, thanks!

PR #128, opened 2022-07-01
sysutils/fusefs-ntfs: Support device offset to mount

kgermanov:
Usually NTFS volume located on device at some offset.
We can use losetup/mdconfig for map device from offset to another device and mount this device.
But there are some problems:

It is new resourse, which has limited && can be leak
Method of mount is not portabled
Need copy data twice: for loop device and fuse
If we pass argument for device offset we can avoid such problems.


grahamperrin:
Context: FreeBSD bug [264983 – sysutils/fusefs-ntfs: Support device offset to mount](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=264983)

rosorio:
Submitted as [PR](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=264983), closed (not accepted).

PR #133, opened 2022-10-14
Add support to build ports as PIE executables. 

jatinkataria:
Extend the current mechanism to provide an option for the ports to compile as PIE executable.
This will allow utilization of ASLR provided by the kernel and improves the cost of the attacker to do control flow hijacking of services.

Sponsored by:Netflix

### DIFF
--- a/Mk/Features/pie.mk
+++ b/Mk/Features/pie.mk
@@ -1,0 +1,14 @@
+# PIE Support
+
+.if !defined(_PIE_MK_INCLUDED)
+_PIE_MK_INCLUDED=	yes
+PIE_Include_MAINTAINER=	portmgr@FreeBSD.org
+
+.  if !defined(PIE_UNSAFE)
+PIE_CFLAGS?=	-fPIE -fPIC
+CFLAGS+=	${PIE_CFLAGS}
+CXXFLAGS+=	${PIE_CFLAGS}
+LDFLAGS+=	-pie
+.  endif
+.endif
+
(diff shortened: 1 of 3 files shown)


Mk/Features/pie.mk
```diff
@@ -0,0 +1,14 @@
...
+
+.if !defined(_PIE_MK_INCLUDED)
+_PIE_MK_INCLUDED=	yes
+PIE_Include_MAINTAINER=	portmgr@FreeBSD.org
+
+.  if !defined(PIE_UNSAFE)
+PIE_CFLAGS?=	-fPIE -fPIC
+CFLAGS+=	${PIE_CFLAGS}
+CXXFLAGS+=	${PIE_CFLAGS}
+LDFLAGS+=	-Wl,--as-needed -pie -Wl,-z,relro -Wl,-z,now
```
bapt, line 11: LDFLAGS do contain way more then needed for pie. --as-needed might be a good global option, but will have breakage for sure and should not belong to PIE. same for relro (I don't expect failure for that) and for  -z,now.

jatinkataria: Thanks for that feedback and I will refactor them. I can create separate feature files for those flags. Is there a recommendation to have one file per flag or a "elf_hardening.mk" would be ok to include all the relevant flags?

bapt: relro and bindnow can be merged into one separate mk like elf_hardening yes.

bapt:
merged and I also added the RELRO and BIND_NOW features. I also amended the commit log, I hope you are ok with the changes

PR #134, opened 2022-10-18
UPDATING: Fix some typos

ykla:
According to https://cgit.freebsd.org/ports/commit/?id=e025129daf0d7006391e5311f20ec6102fb3d889 , the date of this line should be 2021 not 2020.


metalefty:
Good catch.

PR #135, opened 2022-11-09
Fix typo in ircd-hybrid description

yonas:
Change "irc" to "IRC".


yonas:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=267669

PR #139, opened 2022-12-09
Update to multimedia/py-soco to v0.28.1, which was released on 2022-10-22

conallob:


metalefty:
Pushed, thanks. You don't need to make a new pull request but just force-push'ing to your previous pull request is fine.
Anyway, thanks for your work!

PR #144, opened 2023-01-17
add Homebridge as uid 866

ohmantics:
Pretty happily running [Homebridge](https://github.com/homebridge/homebridge) in an iocage-based jail. Homebridge accepted my patches for FreeBSD support last year and install instructions are like [this](https://gist.github.com/ohmantics/11516438b693a91e8bca4edcba7f796b).


arrowd:
UIDs lists identifiers used by ports. If there is no port for Homebridge, there is no point in having this entry.

0mp:
@arrowd is correct. Let's add the user to the UIDs file when the Homebridge port lands.

PR #145, opened 2023-01-17
New port: Kubernetes API server 1.26.0 (+ better descr on scheduler & controller-manager)

nerzhul:


grahamperrin:
Cross-reference FreeBSD bug [267308 – [New Port] kube-apiserver](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=267308)

PR #150, opened 2023-01-26
multimedia/emby-server(-devel): pass -P to daemon(8)

jcassette:
With the -r option, the -P option must be used instead of the -p option. Otherwise, when stopping the service, daemon will restart the child. See man daemon(8).

### DIFF
--- a/multimedia/emby-server-devel/Makefile
+++ b/multimedia/emby-server-devel/Makefile
@@ -1,5 +1,6 @@
 PORTNAME=	emby-server
 DISTVERSION=	4.8.0.21
+PORTREVISION=	1
 CATEGORIES=	multimedia
 MASTER_SITES=	https://github.com/MediaBrowser/Emby.Releases/releases/download/${DISTVERSION}/ \
 		https://mediabrowser.github.io/embytools/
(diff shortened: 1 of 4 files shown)


driesmp:
I'll take this

driesmp:
I don't think switch -p to -P is enough, when querying status of the daemon with eg service emby-server status, it is not detected running when it is. The same for stopping the daemon. Have you given your patch a shot?

I think both are needed with -r probably something like here can be adopted, creating a pid dir, where both the pid of the child and the daemon are located.

https://cgit.freebsd.org/ports/tree/net-p2p/radarr/files/radarr.in

driesmp:
So weird, I can't get it to work on my setup with -P or even with -P and -p.

ConstrictM:
The short answer if you need both a daemon pidfile and pidfile.

Also the services need to be killed as they do not shutdown gracefully.

driesmp:
Oh found it, do you have an e-mail address jcassette?

driesmp:
> You can also push to my branch since "Allow edits by maintainers" is enabled
Doesn't matter, as we cant commit from GH, its read only.

driesmp:
Should be fixed!

PR #152, opened 2023-01-28
net/storj: add an rc script for storagenode

jcassette:
This creates a user, folder and rc script to run storagenode as a service.

### DIFF
--- a/GIDs
+++ b/GIDs
@@ -809,7 +809,7 @@ opensearch:*:855:
 # free: 865
 # free: 866
 # free: 867
-# free: 868
+storj:*:868:
 keyd:*:869:
 buildbot:*:870:
 # free: 871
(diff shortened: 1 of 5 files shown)


jbeich:
Requires review/approval by @danfe (maintainer). Looks like this hasn't been submitted via bugzilla, so cannot land under "maintainer timeout".

danfe:
> Do you suggest that I submit this via Bugzilla?

I suggest you maintain the script locally if you find it useful.  Storj typically requires careful setup which is different between users.  E.g. you mention separate `--confir-dir` and `--identity-dir` which are usually kept together, but do not mention locations for metadata (which is often moved to SSD for better speeds) and data (usually resides on a separate storage space).

0mp:
I opened a PR on Bugzilla, so that is it easier for the maintainer to notice this patch:

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=281930

diizzyy:
Closing this as we don't need to track patches on multiple sites

PR #153, opened 2023-01-31
Update website URL and license for py-rope

lieryan:
Hi, I am the maintainer of python-rope, just updating this outdated metadata here.

Rope has moved from Sourceforge to GitHub and has [relicensed to LGPLv3+](https://github.com/python-rope/rope/issues/262).

Also noted that rope 0.18.0 is severely out of date, the latest version of rope is 1.7.0. I don't use BSD, so I can't really test it on this OS, so I am leaving them as it is; but if anyone is actually interested in installing this package from BSD Ports, it would be appreciated to update the version here as well.

### DIFF
--- a/devel/py-rope/Makefile
+++ b/devel/py-rope/Makefile
@@ -6,9 +6,9 @@ PKGNAMEPREFIX=	${PYTHON_PKGNAMEPREFIX}
 
 MAINTAINER=	jjachuf@gmail.com
 COMMENT=	Python refactoring library
-WWW=		http://rope.sourceforge.net
+WWW=		https://github.com/python-rope/rope/
 
-LICENSE=	GPLv2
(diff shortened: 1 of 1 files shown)


grahamperrin:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=273865

lieryan:
@grahamperrin thanks for looking into this ticket. The discussion by the various authors for the relicensing takes place in https://github.com/python-rope/rope/issues/262. As far as I can tell, the intent was to relicense the project to LGPL3+. 

Please let me know if you think there's anything that needed to be fixed on rope's side to comply with any FreeBSD policies in regards to the license notices.

arrowd:
This seems to be merged, closing.

PR #159, opened 2023-02-15
multimedia/emby-server(-devel): create daemon pidfile

ConstrictM:
Finishes #150 and also properly kills the daemon and child services.

@driesmp


driesmp:
Should be fixed in 6e69b81

PR #161, opened 2023-02-17
multimedia/emby-server(-devel): remove redundant pid operations

ConstrictM:
@driesmp 

Something still to note here:
The rc.subr automatically creates and removes the pid file.   "-- /dev/null" in the install hides that there is an error creating a file that already exists.

### DIFF
--- a/multimedia/emby-server-devel/Makefile
+++ b/multimedia/emby-server-devel/Makefile
@@ -1,6 +1,6 @@
 PORTNAME=	emby-server
 DISTVERSION=	4.8.0.21
-PORTREVISION=	2
+PORTREVISION=	3
 CATEGORIES=	multimedia
 MASTER_SITES=	https://github.com/MediaBrowser/Emby.Releases/releases/download/${DISTVERSION}/ \
 		https://mediabrowser.github.io/embytools/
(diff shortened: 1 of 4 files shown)


driesmp:
Redundant operations is a good improvement! => Can you update this PR to only include that commit?

Disabling Vulkan is meh, in poudriere it builds fine as it doesn't get poisoned by headers. 
Either we leave it as is (which I like), or we disable all autodetected features like Vulkan.

ConstrictM:
done @driesmp

driesmp:
Great, do you have an e-mail address I can use? And optionally your full name including surname. If you don't want to disclose I can use your github username.

driesmp:
The patch has errors in starting Emby though. When removing the creation it will try install a PID file as emby, but it can't write in /var/run. If we want to get rid of the things in the PR, we'll have to add a /var/run/emby-server dir owned by emby:emby, and move the pid file there.

ConstrictM:
Not entirely sure how I missed that...must have some wrong permissions in my instance.

Added the directory and set the permissions.

Matthew Wener
matthew@wener.org

driesmp:
Committed! Thanks

PR #166, opened 2023-03-09
Enable load of tpm driver with devd rule

nangirl:
The devd rule comes from @tstruk proposal. The old rule was not working. This rule won't trigger if tpm driver was already loaded at boot.

### DIFF
--- a/security/tpm2-abrmd/files/tpm2-abrmd.in
+++ b/security/tpm2-abrmd/files/tpm2-abrmd.in
@@ -41,6 +41,11 @@ tpm2_abrmd_start()
     touch $logfile
     chmod 640 $logfile
     chown -R ${tpm2_abrmd_user}:${tpm2_abrmd_group} ${logfile}
+    
+    if ! kldstat -d -q -n tpm.ko; then
+        kldload tpm
+    fi
+    
     /usr/sbin/daemon -P ${pidfile} -u ${tpm2_abrmd_user} ${command} >>${logfile} 2>&1
 
     if is_process_running; then


tstruk:
Looks good to me. Thanks!

LuKePicci:
@bsdimp devfs doesn't fit well to the "packaged" rule addition approach. AFAIK there's no way to include additional devfs rules by adding a file. The only devfs way would be to define such rule into something which is not part f this package. Does the _tss group exist before installing this port? Would it be fine to have devfs setting permissions to a non-existing group id?

BTW, I made a quick test on pfSense with devfs.rules by setting a new system default ruleset and extending the previous default ruleset with a new extra rule for tpm.  It didn't work until I restarted devfs despite tpm.ko was correctly loaded during boot with tpm_load="YES". Not sure if pfSense is doing strange things with devfs or this tpm driver couldn't work with devfs.

LuKePicci:
@puffetto I'm reviewing your latest changes to the tpm2-abrmd port. I have to point to let you consider:
1. The previous devd conf you removed was wrong, the corrected version is in this PR #166 
2. This PR was intended to allow the tpm driver to be kldloaded after boot (as we use this approach in a hundred of FreeBSD-based gateways in production on our customer sites for more then 2 years now). At that time, we did test loading the driver at boot time and applying devfs rules, but it didn't work.

I can help you testing this in deep as we use this driver for IoT Hub TPM-based attestation since a long time. Feel free to contact me also on telegram (same nickname as gh) or come visit us in our offices in Milan. 😅

PR #168, opened 2023-03-14
Added support for RTL8125 labelled as Killer E3000/E3100

elgithubo:


alexdupre:
Committed to upstream branch, thanks.

PR #171, opened 2023-04-02
www/nginx: start only with nginx_checkconfig success

kulikov-a:
Hi!
Is it intentionall that `nginx_prestart()` ignores the `nginx_checkconfig` result?
So it is possible that nginx will try to start with a known broken config? (which can lead, for example, to the prevention of subsequent starts if the `listen` socket is not deleted on failed startup).
Thanks!

### DIFF
--- a/www/nginx/files/nginx.in
+++ b/www/nginx/files/nginx.in
@@ -140,7 +140,7 @@ nginx_prestart()
 		required_modules="$required_modules accf_http accf_data"
 	fi
 
-	nginx_checkconfig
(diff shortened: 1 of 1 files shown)


jbeich:
Requires review/approval by @joneum (maintainer). Looks like this hasn't been submitted via bugzilla, so cannot land under "maintainer timeout".

rosorio:
PR created: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=276659

PR #173, opened 2023-04-17
net-mgmt/unbound_exporter: fix typos in rc.d script

wblew:
There are typos in two arguments to unbound_exporter within its service script. This fixes typos in the unbound_exporter_bind and unbound_exporter_path variables application to unbound_exporter startup.

They prevent successful startup of unbound_exporter with `service unbound start`. The unbound_exporter process fails to
start due to errors in its command arguments.

This PR addresses those typos. It has been tested on my pfSense appliance.

PS: I am unclear whether this should be submitted here as a PR, or as a patch in another manner somewhere else. If you need
anything changed, just let me now!


wblew:
I have also updated a relevant Bugzilla entry with this patch:

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=270442

PR #178, opened 2023-05-19
devel/rubygem-asetus: Update to 0.4.0

lassebm:
Fixes `rubygem-oxidized` failing to start with Ruby 3.1, see https://github.com/ytti/oxidized/issues/2730.


jbeich:
Requires review/approval by @nickhilliard (maintainer). Looks like this hasn't been submitted via bugzilla, so cannot land under "maintainer timeout".

nickhilliard:
submitted on bugzilla via https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=271542

@lassebm the freebsd gh repo is a read-only mirror which doesn't accept pull requests. If you want to submit patches, etc, they need to go into bugzilla.freebsd.org.

PR #181, opened 2023-06-20
net-mgmt/bsnmp-ucd: Fix WWW

tomhukins:


yurivict:
Patch is committed.
Thanks for the patch.

PR #184, opened 2023-07-16
games/melonds: Update to 0.9.5

nadiaholmquist:
This updates the melonDS emulator to 0.9.5, which is the latest stable version.

This required backporting a patch which fixes compilation with newer versions of clang, as well as backporting another patch that I just made upstream that fixes a strange issue with the Qt 5 flavor crashing immediately on launch on some configurations, notably FreeBSD and apparently also some Linux systems.

This is my first contribution to ports so I hope I've done everything right, and if there's something upstream could do to make it easier to package for BSD, please do let me know :)

Edit: force pushed an amended commit because I realized I forgot to update the dependencies.

### DIFF
--- a/games/melonds/Makefile
+++ b/games/melonds/Makefile
@@ -1,6 +1,5 @@
 PORTNAME=	melonds
 PORTVERSION=	0.9.5
-PORTREVISION=	0
 CATEGORIES=	games
 PKGNAMESUFFIX=	-${FLAVOR}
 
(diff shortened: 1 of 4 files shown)


lwhsu:
cc maintainer @HenryHu

games/melonds/Makefile
```diff
@@ -1,6 +1,6 @@
 PORTNAME=	melonds
-PORTVERSION=	0.9.4
-PORTREVISION=	1
+PORTVERSION=	0.9.5
+PORTREVISION=	0
```
lwhsu, line 3: No need to reset to 0, just remove this line.

lwhsu:
After testing this in the poudriere, it still needs:
  `USES= kde:5`
and
  `USE_KDE= ecm:build`

nadiaholmquist:
That should be those two changes taken care of. However I just remembered that 0.9.5, because of its custom OpenGL workarounds, might need to depend on X11 and Wayland libraries directly.

I'll have to try and set up `poudriere` to make sure I've got this all right.

diizzyy:
Hi,

Please submit this to https://bugs.freebsd.org since we cannot land this as we don't officially support this repo as a way of submitting patches to the ports tree.

Best regards,
Daniel

lwhsu:
merged as 3e10ec461ae5e04573c6180230022e1ccb809f15

PR #186, opened 2023-07-20
misc/tzdiff: Reorder variables.

yuichiro-naito:
Reorder variables in order to FreeBSD Porters Handbook. This is the reflection of [the comment in FreeBSD bugzilla](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=272610#c1).

1.  Put USE_GITHUB upper on GH_ACCOUNT.
2. Put PLIST_FILES below NO_ARCH & NO_BUILD.

### DIFF
--- a/misc/tzdiff/Makefile
+++ b/misc/tzdiff/Makefile
@@ -8,14 +8,13 @@ WWW=		https://github.com/belgianbeer/tzdiff
 
 LICENSE=	BSD2CLAUSE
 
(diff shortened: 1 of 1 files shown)


diizzyy:
Fixed in https://cgit.freebsd.org/ports/commit/misc/tzdiff?id=99af3eef26c5b97f692bc3c56a62509d65d92d38
Thanks! :-)

PR #194, opened 2023-10-08
dict: Update pkg-descr

chrislongros:
More detailed info about the purpose and mission of dict


chrislongros:
@wosch

rosorio:
Thanks Christos this description is much better than the existing one
specially after the removal of the WWW leaving an empty sentence.
Unfortunately the maintainer stop updating the file in 1999 which means
this port really has 114842 entries.

I look a little bit forward about how to update the port itself, but it seems
that the license terms from the project are incompatible the way ports works.

https://www1.dict.cc/translation_file_request.php?l=e

* we can't repackage the translation database it has to be downloaded
   by the users app from dict.cc
* the application who uses the data has to comply with GPL

In the current paort we package the archive and the script used to retrieve translation is une BSD licence.

rosorio:
Are you currently using this port ?

chrislongros:
> Are you currently using this port ?

I contacted the owner of dict.cc. Downloading the dictionary database is not approved. I try to implement a python script that fetches and queries the online dictionary without the need of a local database.

rosorio:
Since we can't update neither this port or the description, we can close this PR.
If you are working in an alternative using the online data, I suggest you to submit
a PR in our [Bugzilla](bugs.freebsd.org) to add this new port and request deprecation of this one.

PR #195, opened 2023-10-17
www/falkon: update package description

grahamperrin:
It's no longer 'new-ish'.

Whilst here, rewrite things slightly.

### DIFF
--- a/www/falkon/pkg-descr
+++ b/www/falkon/pkg-descr
@@ -1,17 +1,17 @@
-Falkon is a new-ish and very fast QtWebEngine browser. It aims to be a
-lightweight web browser available on all major platforms. Falkon
-has all standard functions you expect from a web browser. It includes
-bookmarks, history (both also in sidebar) and tabs. Beyond that, you
-block ads with a builtin AdBlock plugin, block Flash content with
-Click2Flash and edit the local CA Certificates database with an SSL Manager.
+Falkon is a very fast web browser that uses QtWebEngine, and aims to be
+lightweight on all major platforms.  It has all the standard functions
+that are expected from a web browser, including bookmarks, history,
+sidebar, and tabs.  Plus: integral AdBlock, and edition of the local CA
+Certificates database with an SSL Manager.
 
-Falkon has a wide range of features supported by KDE Frameworks
-and many integration points with KDE Plasma. A "-qtonly" flavor is
-available without these features:
+Falkon has a wide range of features that are supported by KDE
+Frameworks, and many integration points with KDE Plasma.
+
+The "-qtonly" flavor omits these features:
 - spellchecking
 - word completion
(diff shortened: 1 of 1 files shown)


grahamperrin:
Background: [#37 - FreeBSD Handbook: consistency, modernisation: www/falkon - freebsd-doc - Codeberg.org](https://codeberg.org/grahamperrin/freebsd-doc/pulls/37)

grahamperrin:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=274547

BSDKaffee:
We generally avoid using double spaces after a full stop. The use of a single space is standard in modern writing. "Edition" is a noun. I would replace it with the verb "editing". The word "Manager" in "SSL Manager" should be lowercase since it used generically. Bulleted lists that don't complete the sentence stem should not be punctuated. I would also remove the hyphen before "qtonly" because that is the name of the flavor.

grahamperrin:
Thank you, 

> The word "Manager" in "SSL Manager" should be lowercase since it used generically. 

Should the same be done for the _CA certificates database_? There's currently an uppercase C for Certificates.

grahamperrin:
> We generally avoid using double spaces after a full stop. The use of a single space is standard in modern writing. 

Agreed, generally; with exceptions in some FreeBSD Project contexts – and the exceptions are not yet suitably documented. My https://github.com/freebsd/freebsd-src/commit/edacf4b4824af3630a953eb1080baf9c46f176de (removal of a double space) was treated as an error, and so on. 

Briefly: can you think of any context where representation of a package description, making _direct_ use of e.g. the `pkg-descr` file for a port, is **not** monospaced?

BSDKaffee:
> Should the same be done for the _CA certificates database_? There's currently an uppercase C for Certificates.

Yes, it is a generic term. I'd make it singular too. "CA certificate database" sounds better.

BSDKaffee:
> Briefly: can you think of any context where representation of a package description, making _direct_ use of e.g. the `pkg-descr` file for a port, is **not** monospaced?

Probably any port maintained by danfe@. He's a fellow grammar pedant, which I appreciate, but I disagree about double spaces after full stops. It's not wrong, just a dated writing style. It used to write that way myself, to be honest, but have kicked the habit.

BSDKaffee:
> My [freebsd/freebsd-src@edacf4b](https://github.com/freebsd/freebsd-src/commit/edacf4b4824af3630a953eb1080baf9c46f176de) (removal of a double space) was treated as an error, and so on.

I don't know why removing a double space from UPDATING would cause an error. There is a great lack of consistency with regards to double space vs. monospace after a full stop in that file, even within some of the individual entries.

arrowd:
Merged, thanks.

PR #198, opened 2023-11-15
net/sacc: remove me from maintainer

c0dev0id:
Hi,
please remove me as maintainer from net/sacc. It's a simple port for anyone who wants to take it. I don't have a FreeBSD box anymore.


metalefty:
MAINTAINER section must not be removed from Makefile. It should be set to ports@FreeBSD.org instead but thanks!

PR #201, opened 2023-12-06
Update Jellyfin version

rucko24:
Hello very good morning, 

I understand that we can already use the latest release, with the patching of the input with **FFmpeg**.


rosorio:
Hi @rucko24,

Thanks for the pull request, but unfortunately someone report the it on our bugzilla a month ago with a more elaborate patch.
I encourage you to have a look at the submitted patch and see what's missing in yours. We have a manual guide about porting who explains all the requirements and tricks to successfully update a port: https://docs.freebsd.org/en/books/porters-handbook/ .

cheers
- rodrigo

PR #203, opened 2023-12-17
Update net/libsockpp to 0.8.3, add devel/libasyncplusplus port

Peter2121:


arrowd:
Merged, thanks.

I had to remove `STATIC` option in `devel/libasyncplusplus`, because defaulting to installing only a static library is a no-go. Try reaching out upstream and ask them to support building the library in two ways simultaneously.

PR #217, opened 2024-01-18
converters/hs-aeson-pretty: update to 0.8.10

Awoonyaa:
@arrowd


arrowd:
Much appreciated, updating git-annex is pain in the ass.

PR #223, opened 2024-01-25
net/miniupnpd: Update package description to include PCP as supported

Self-Hosting-Group:
Port Control Protocol (PCP) is the successor to NAT-PMP and has similar protocol concepts and packet formats, but adds IPv6 support.

*PCP standard:*
https://datatracker.ietf.org/doc/html/rfc6887
https://en.wikipedia.org/wiki/Port_Control_Protocol
*NAT-PMP std. (see 9.1 Simplicity - 9.3 for some diff. to UPnP IGD):*
https://datatracker.ietf.org/doc/html/rfc6886

### DIFF
--- a/net/miniupnpd/Makefile
+++ b/net/miniupnpd/Makefile
@@ -4,7 +4,7 @@ PORTEPOCH=	1
 CATEGORIES=	net
 
 MAINTAINER=	squat@squat.no
-COMMENT=	UPnP IGD implementation which uses pf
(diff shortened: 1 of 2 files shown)


rosorio:
reported as https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=276657

arrowd:
https://downforeveryoneorjustme.com/miniupnp.tuxfamily.org reports that new WWW doesn't work.

arrowd:
Pushed, thanks.

PR #225, opened 2024-01-28
[NEW PORT] graphics/entice Simple Image Viewer Using EFL

Peter2121:


stesser:
This port is not fully compliant with FreeBSD conventions (see the Porter's Handbook). 
I could easily fix them, but I'd think it would be better if you applied the (minor) corrections yourself:

- We try to avoid using a GitHub commit hash, only tags should be used, if possible at all. (You may ask the author of entice, whether at least at tag like 0.0.1 could be applied).
- When using a date as the version, the PORTVERSION should be specified and it should start with the latter "g". this will make any such version sort lower than any numeric version (even 0.0), if numeric version tags are introduced at a later time.
- Ports should be checked at least with portlint, but better also with portclippy and portfmt (see the ports-mgmt/portfmt port). The Makefile has incorrect indentation which will be fixed by portfmt.
- There is a misspelling of "Enlightenment" (missing "n") in pkg-descr (and that line should end with ".").

If these details are fixed I'd be willing to commit the port (after having checked that "poudriere testport" finds no further issues).

Peter2121:
Hi Stefan,

Thanks for your suggestions. 
As about tag - no chance, I've asked the author already.
I'll fix another points ASAP.

Best regards,
Peter

stesser:
Hi Peter,
you missed "make makesum" after changing the PORTVERSION (which leads to a different distfile name with unmodified contents). A build test with "poudriere testport" failed since the dependency "efl" could not be built with default options, but a build test in a clean jail (and efl pre-built without the broken Wayland-support) succeeded.
Thank you for your contribution.

0mp:
Hey Peter, it seems like the port is almost ready to land. The only missing thing is to run `make makesum` to regenerate the distinfo file.

If you'd like to continue working on this port, please let us know. In the meantime, I'll close this PR.

PR #232, opened 2024-02-08
sysutils/tt: New port

0x501D:
Command-line utility to manage Tarantool applications.

### DIFF
--- a/sysutils/tt/Makefile
+++ b/sysutils/tt/Makefile
@@ -1,0 +1,127 @@
+PORTNAME=	tt
+PORTVERSION=	2.1.2
+DISTVERSIONSUFFIX=	-complete
+MASTER_SITES=	https://github.com/tarantool/tt/releases/download/v${PORTVERSION}/
+WRKSRC=         ${WRKDIR}/${PORTNAME}-${PORTVERSION}
+CATEGORIES=	sysutils databases
+
+MAINTAINER=	mail@void.so
+COMMENT=	Command-line utility to manage Tarantool applications
+WWW=		https://github.com/tarantool/tt
+
+LICENSE=	BSD2CLAUSE
+LICENSE_FILE=	${WRKSRC}/LICENSE
+
+ONLY_FOR_ARCHS=		amd64 aarch64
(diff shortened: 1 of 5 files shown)


arrowd:
@wahjava You're knowledgeable in Go, would mind looking at this?

sysutils/tt/Makefile
```diff
@@ -0,0 +1,127 @@
...
+BUILD_DEPENDS=  ${LOCALBASE}/bin/mage:devel/mage
+
+USES=		go:modules
+
+PLIST_FILES=	bin/tt
+
+do-build:
+	@cd ${WRKSRC} && ${SETENV} ${MAKE_ENV} ${GO_ENV} ${LOCALBASE}/bin/mage build
+
+GH_TUPLE=	\
```
metalefty, line 27: Now we can use GO_MODULE instead of GH_TUPLE.

wahjava: Yep, it'll also help with fetching modules which aren't on GitHub.

wahjava (changes requested):

sysutils/tt/Makefile
0x501D: Unfortunately, tt does not work as a go module due to the complexity and structure of the project itself:
```sh
$ go get github.com/tarantool/tt@latest
go: github.com/tarantool/cartridge-cli@v0.0.0-20220605082730-53e6a5be9a61 (replaced by ./cli/cartridge/third_party/cartridge-cli): reading cli/cartridge/third_party/cartridge-cli/go.mod: open /home/void/tmp/cli/cartridge/third_party/cartridge-cli/go.mod: no such file or directory
```

0x501D: Is it possible to just remove `USES= go:modules`, GH_TUPLE and just do the mage build in the Makefile?

0x501D: And if port absolutely need to build without a network, I can try to use a cache of vendor libraries, the `tt` repository has them for `tt` gentoo ebuild.

PR #234, opened 2024-02-16
x11/gbsddialog: add new port

khorben:
gbsddialog is an implementation of bsddialog, a text-based user interface to be used in scripts and tools, such as in the official FreeBSD installer, bsdinstall(8). gbsddialog uses Gtk+ dialogs and widgets instead of console-based equivalents. The objective is to reproduce the same behaviour as the original tool, respecting the exact command-line parameters and output. gbsddialog can then be used as a drop-in replacement to bsddialog, effectively granting the possibility to install and configure FreeBSD entirely in graphical mode.

In addition, an implementation of Xdialog is also provided, for use with bsdconfig(8) when invoked in graphical mode with -S or -X.

Sponsored by: The FreeBSD Foundation


lwhsu:
The `distinfo` file is missing, you can create it via `make makesum`.

lwhsu:
This needs libbsddialog, so we need to add:

`LIB_DEPENDS=    libbsddialog.so:devel/bsddialog`

lwhsu:
After applying above fixes, it still cannot be built in poudriere: https://gist.github.com/lwhsu/2a46f151b227ad899052a51476b519c3

lwhsu (changes requested):
Please address the comments.

khorben:
> The `distinfo` file is missing, you can create it via `make makesum`.

Yes I forgot to `git add` it; I have updated the pull-request already.

I am looking into the other issues you noted now.

Thank you for the feedback!

khorben:
> This needs libbsddialog, so we need to add:
> 
> `LIB_DEPENDS= libbsddialog.so:devel/bsddialog`

Actually it does not need the library file, it only uses `<bsddialog.h>` at compile-time for consistency with `bsddialog`. But I see now that `bsddialog.h` in base has been made private, so I will probably re-release gbsddialog as 0.8.1, after adding this file.

This should also avoid the compatibility issue found in the poudriere build.

khorben:
Superseding with a pull-up request from `khorben:khorben/gbsddialog-0.8.1`.

PR #237, opened 2024-02-22
Add security/xhash

ricardobranco777:
Add security/xhash

This Go program uses goroutines to calculate multiple hashes on strings, files and directories.  By default it reads from standard input.  It can be used as a drop-in replacement for the GNU **coreutils** when hard-linked as **md5sum**, etc. and it actually supports the `--zero` option with `--check`.  The output format is fully configurable.

### DIFF
--- a/security/xhash/Makefile
+++ b/security/xhash/Makefile
@@ -1,0 +1,26 @@
+PORTNAME=	xhash
+DISTVERSIONPREFIX=	v
+DISTVERSION=	3.3.1
+CATEGORIES=	security
+
+MAINTAINER=	rbranco@suse.de
+COMMENT=	Calculate hashes on files using goroutines
+WWW=		https://github.com/ricardobranco777/xhash/
+
+LICENSE=	MIT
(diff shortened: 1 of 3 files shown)


metalefty:
Have you tested the build? `USES= go:modules,1.21` is required.

```
===>  Building xhash from .
check.go:9:2: package slices is not in GOROOT (/usr/local/go120/src/slices)
note: imported by a module that requires go 1.21
*** Error code 1

Stop.
make: stopped in /usr/ports/security/xhash
build of security/xhash | xhash-3.3.1 ended at Sat Feb 24 22:47:52 JST 2024
build time: 00:00:02
!!! build failure encountered !!!

```

ricardobranco777:
> Have you tested the build? `USES= go:modules,1.21` is required.

Fixed, thanks.

I tested it but I had `DEFAULT_VERSIONS+= go=1.22` set in /etc/make.conf.  It works as well with go1.21.

metalefty:
Thanks for your contribution!

PR #239, opened 2024-02-24
graphics/plantuml: added an ability to pass command line flags

Awoonyaa:
@arrowd


arrowd:
Pushed, thank you.

VVD:
@Awoonyaa, https://bugs.freebsd.org/295670

PR #240, opened 2024-02-27
Update security/xhash

ricardobranco777:
Fix 32-bit compilation and add support for BLAKE2s


Zirias:
I'll give it some quick build tests, thanks.

PR #241, opened 2024-03-03
Port addition: aws-sam-cli

ElectricPulse:
Addition of the port `aws-sam-cli` which I would actively like to use for work purposes.

### DIFF
--- a/devel/aws-sam-cli/Makefile
+++ b/devel/aws-sam-cli/Makefile
@@ -1,0 +1,42 @@
+PORTNAME=	aws-sam-cli
+PORTVERSION=	1.110.0
+PORTREVISION=	1.0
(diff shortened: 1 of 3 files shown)


devel/aws-sam-cli/Makefile
```diff
@@ -0,0 +1,43 @@
+PORTNAME=	aws-sam-cli
+PORTVERSION=	1.110.0
+PORTREVISION=	1.0
```
0mp, line 3: PORTREVISION must not have to be set for new ports.

0mp (changes requested):
Hey! Thanks for the patch.

Have you tested if the ports builds in poudriere?

PR #246, opened 2024-04-11
Uses/electronfix.mk: Add electron28 to _ELECTRONFIX_MK_VALID_VERSIONS

Zoid-dr:


Mk/Uses/electronfix.mk
```diff
@@ -17,7 +17,7 @@
 .if !defined(_INCLUDE_USES_ELECTRONFIX_MK)
 _INCLUDE_USES_ELECTRONFIX_MK=	yes
 
-_ELECTRONFIX_MK_VALID_VERSIONS=	22 23 24 25
+_ELECTRONFIX_MK_VALID_VERSIONS=	22 23 24 25 28
```
tagattie, line 20: Currently, the ports tree has 3 versions of electron, namely 25, 27, and 28.

So this line should look like:

```
_ELECTRONFIX_MK_VALID_VERSIONS=	25 27 28
```

tagattie (changes requested):

PR #247, opened 2024-04-12
security/vuxml: make extra-validation.py work again

khorben:
This script probably stopped working since 87748de, as Python's xml.etree.ElementTree module is not capable of expanding external entities. (As per https://docs.python.org/3/library/xml.html)

This change makes use of the python-lxml module instead.

Sponsored by:	The FreeBSD Foundation

### DIFF
--- a/security/vuxml/files/extra-validation.py
+++ b/security/vuxml/files/extra-validation.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 
 import datetime
-import xml.etree.ElementTree as ET
+from lxml import etree as ET
 import sys
 import re
 
@@ -15,7 +15,8 @@ re_invalid_package_name = re.compile('[@!#$%^&*()<>?/\|}{~:]')
 # warn if description has more than X characters
 DESCRIPTION_LENGTH = 5000
 
-tree = ET.parse(sys.argv[1])
+parser = ET.XMLParser(dtd_validation=True)
+tree = ET.parse(sys.argv[1], parser)
 root = tree.getroot()
 
 namespace = "{http://www.vuxml.org/apps/vuxml-1}"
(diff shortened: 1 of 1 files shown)


Jehops:
I we do indeed need to depend on lxml module, we would have to add something like `${PYTHON_PKGNAMEPREFIX}lxml>0:devel/py-lxml@${PY_FLAVOR}` to `RUN_DPENDS`.  We would also have to bump `PORTREVISION`.

```
% make validate
xmllint -noent /usr/ports/security/vuxml/vuln.xml > /usr/ports/security/vuxml/vuln-flat.xml
/bin/sh /usr/ports/security/vuxml/files/tidy.sh "/usr/ports/security/vuxml/files/tidy.xsl" "/usr/ports/security/vuxml/vuln-flat.xml" > "/usr/ports/security/vuxml/vuln.xml.tidy"
>>> Validating...
/usr/local/bin/xmllint --valid --noout /usr/ports/security/vuxml/vuln-flat.xml
warning : xmlAddEntity: invalid redeclaration of predefined entity 'lt'
warning : xmlAddEntity: invalid redeclaration of predefined entity 'amp'
>>> Successful.
Checking if tidy differs...
... seems okay
Checking for space/tab...
... seems okay
/usr/local/bin/python3.9 /usr/ports/security/vuxml/files/extra-validation.py /usr/ports/security/vuxml/vuln-flat.xml
Traceback (most recent call last):
  File "/usr/ports/security/vuxml/files/extra-validation.py", line 4, in <module>
    from lxml import etree as ET
ModuleNotFoundError: No module named 'lxml'
*** Error code 1

Stop.
make: stopped in /usr/ports/security/vuxml
```

khorben:
Sorry, I did not see that `make validate` expands the XML data to a flat file in `make do-test` before calling `extra-validation.py`. Since the script is not shipped, I thought it was exclusively used manually out of tree, and that therefore there was no point in registering additional dependencies in the `Makefile`.
I guess you mean the dependency should be added for build-time and not run-time?
Anyway, all good, please ignore this Pull Request then.

PR #251, opened 2024-04-23
fusefs-lkl: gcc+binutils is only needed to build

motminh:
There are no run-time requirements on GCC, so don't force the install of GCC, and worse Binutils to conflict with base utilities.


asomers (changes requested):
You should increment PORTREVISION to force a new package to be built.

PR #254, opened 2024-04-30
games/xmoto: Fix database double-quoting crash

vasi:
When sqlite is built with DQS disabled by default, xmoto still needs this SQL quirk.

See https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=278669


jbeich:
Obsolete after 52371ce420c7? If so close the pull request.

PR #263, opened 2024-05-19
security/xhash: Bump to v3.3.3, go122 and latest dependencies

ricardobranco777:
Bump to v3.3.3, go122 and latest dependencies

Ref: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=279158


ricardobranco777:
Addressing in bugzilla.

PR #266, opened 2024-05-22
mail/evolution: update to 3.50.2 (and dependencies)

clausecker:
This patch set was gratefully provided by @arrowd.  It updates mail/evolution and required dependencies, mainly to migrate some stuff to libsoup3.  I have not yet been able to test all dependencies of webkit2-gtk[34], but suspect that more dependents need to be bumped.

This needs to be reviewed by gnome@ most likely.


clausecker:
This fixes most of the API breakage from the webkit2gtk3 update.  However, some cases remain and seem to be mostly unfixable.  The common element is ports that require libsoup2 and actually use the API.  These cannot be updated to webkit2gtk3 API 4.1 as it uses libsoup3 with an incompatible API.  This includes wxgtk3 which has so far not been ported to libsoup3.

Perhaps the mid-term solution is to split webkit2gtk3 into one package for API version 4.0 and one for API version 4.1.  I'll look into that.

arrowd:
All these updates are now merged, so this PR can be closed.

clausecker:
Thank you all for updating these ports.

PR #270, opened 2024-06-04
graphics/mapnik: Python >= 3.10 support

hummeltech:
_Also_:
- Move Boost v1.83 support patch source to GitHub
- Add Boost v1.85 support


hummeltech:
Merged in cc86e8b36a60646a76a6b07843fe6cb427256c0d

PR #277, opened 2024-06-21
security/osv-scanner: Update to 1.8.1

khorben:
Changes:
https://github.com/google/osv-scanner/compare/v1.7.1...v1.8.1

This also fixes detecting if Go is installed locally at run-time, when
checking for known vulnerabilities against source code written in Go.

Sponsored by:	The FreeBSD Foundation

### DIFF
--- a/security/osv-scanner/Makefile
+++ b/security/osv-scanner/Makefile
@@ -1,7 +1,6 @@
 PORTNAME=	osv-scanner
 DISTVERSIONPREFIX=	v
-DISTVERSION=	1.7.1
-PORTREVISION=	3
+DISTVERSION=	1.8.1
 CATEGORIES=	security
 
 MAINTAINER=	lcook@FreeBSD.org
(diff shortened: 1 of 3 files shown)


khorben:
After testing osv-scanner some more, I have found that it checks for the Go compiler to be available locally at run-time, when preparing to analyse source code written in Go. This fails on FreeBSD because it runs "go version", which is typically not working there, since it is missing the version suffix.

I have used inspiration from the security/vuls port and:
* Patched `internal/sourceanalysis/go.go` to call "go%%GO_SUFFIX%% version" instead
* Added a `post-patch` target invoking `${REINPLACE_CMD} -e s,%%GO_SUFFIX%%,${GO_SUFFIX},` for this file

I believe this addresses this issue. It will require the specific version of Go used while building the port to be installed for this feature to work though.

khorben:
Superseded by #282.

PR #281, opened 2024-07-05
Nitrokey and its dependencies 

Awoonyaa:
@arrowd


arrowd:
Let's start with rearranging commits in a correct order so that dependencies are added first.

arrowd:
`comms/py-pyocd` requires `pylink` on runtime:

> ModuleNotFoundError: No module named 'pylink'

arrowd:
Now `comms/py-spsdk` requires updating to its latest version.

PR #283, opened 2024-07-10
devel/llvm18: Backport __stack_chk_guard fix

VexedUXR:
Before this, building a kernel module using clang18 would cause `R_X86_64_REX_GOTPCRELX` to be generated (specifically for `__stack_chk_guard`), which isn't supported by the loader. This was fixed upstream, but it doesn't look like it's making it into llvm18.
Original commit - https://github.com/llvm/llvm-project/commit/26464f2

This issue is also present in llvm14, 15, 16 and 17, but this patch doesn't cleanly apply to any of those. If needed we could do them manually, it's just a one line change after all.

Note that the in-tree clang doesn't have this issue since it's locally patched. See -
https://github.com/freebsd/freebsd-src/blob/main/contrib/llvm-project/llvm/lib/CodeGen/TargetLoweringBase.cpp#L2028

### DIFF
--- a/devel/llvm18/Makefile
+++ b/devel/llvm18/Makefile
@@ -1,6 +1,6 @@
 PORTNAME=	llvm
 DISTVERSION=	18.1.7
-PORTREVISION=	0
+PORTREVISION=	1
 CATEGORIES=	devel lang
 MASTER_SITES=	https://github.com/llvm/llvm-project/releases/download/llvmorg-${DISTVERSION:S/rc/-rc/}/ \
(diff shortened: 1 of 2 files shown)


VexedUXR:
CC @brooksdavis since you're the maintainer.

brooksdavis:
I'll do this somewhat differently (I've been sitting on a an update to 18.1.8 which is pointless for the port since it only changes libc++ which we don't ship.)

VexedUXR:
Closing.
Thanks!

PR #291, opened 2024-08-11
security/pam_fprint, security/fprint_demo: Remove obsolete packages

JohnAZoidberg:
See: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=280749
Both packages have long been deprecated upstream.
They don't work anymore.


lwhsu:
> Do we still need this @lwhsu?

I think we can close them now as there is DEPRECATED flag set. Thanks for the reminding this.

PR #294, opened 2024-08-14
security/openssh-portable: fix missing include for sshd.c

knl:
This patch adds a missing include for `channels.h`. It is needed, as later in the file there is a call to `channel_set_hpn`, which is declared in `channels.h`.

Some ISO C99 compilers may complain about implicit function declaration.


knl:
More context: https://github.com/NixOS/nixpkgs/issues/334154, as nixpkgs pulls the patch from freebsd, and that fails to build on my darwin-aarch64.

knl:
@bdrewery Could you please take a look?

bdrewery:
Committed to the port in commit 7ba88c964b6e5724eec462021d984da3989e6a08. Thanks!

PR #300, opened 2024-09-18
Mk/Uses/proccontrol.mk: add USES=proccontrol

ekhramtsov:
Base has introduced proccontrol(1) in [base@7402f93ef2ea](https://github.com/freebsd/freebsd-src/commit/7402f93ef2ea) to opt-out from execution modes (e.g. protmax after [base@74a1b66cf41d](https://github.com/freebsd/freebsd-src/commit/74a1b66cf41d).

Mk/Uses/proccontrol.mk allows to set process modes for build, e.g. disable protmax and allow wxmap to build e.g. Java when host is restricted (e.g. kern.elf64.allow_wx=0 vm.imply_prot_max=1).

UPD: This is heavily WIP until I get more time to separate out elfctl changes and chase elfctl change in tree. I'll likely publish to Phabricator and close this pull request once done.


ekhramtsov:
See also: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=256477#c20

ekhramtsov:
chrome* IIRC I only build tested for PoC and didn't get to set elfctl on resulting binary, also tested like ~year ago and things could change. I would see/test later.

ekhramtsov:
This had moved to https://reviews.freebsd.org/D47086

elfctl.mk etc.: I would probably push when redone into chain on Phabricator.

PR #304, opened 2024-10-01
Update ttrssd.in - allow ttrssd to run as a specified user

rwv37:
``ttrssd`` (from ``www/tt-rss``) currently runs as the user ``www``, and there doesn't seem to be a way to override that. This change adds an optional flag that can be specified in ``/etc/rc.conf`` to do so.

The flag is ``ttrssd_run_as_user``; for example to run as the user ``mmm``, add the following line to ``/etc/rc.conf``:
```
ttrssd_run_as_user="mmm"
```
If it is not specified, it defaults to ``www`` (reproducing the behavior from before this change).

PLEASE NOTE: This is my first time submitting a change to the FreeBSD ports tree. I tried my changes out on my machine, and they worked as expected. However, when I went to submit my changes here on GitHub, I was surprised to find that there was no file directly the same as what's installed on a machine at ``/usr/local/etc/ttrssd``. Instead, there's this ``ttrssd.in`` file, which has some templating instructions for (I guess) the install process, such as ``%%WWWOWN%%``, which (at least on my machine) is changed to simply ``www`` when installed.

I have altered my original changes to take this into account; in particular changing my added line number 24 to use ``%%WWWOWN%%`` rather than my original ``www``. Unfortunately, I have no idea how to test this, and perhaps I'm misunderstanding something about it, so for anyone who reviews this pull request, please keep an eye out specifically for a mistake I made there. Sorry for the hassle, and thank you.

### DIFF
--- a/www/tt-rss/files/ttrssd.in
+++ b/www/tt-rss/files/ttrssd.in
@@ -7,6 +7,9 @@
 # Add the following lines to /etc/rc.conf to enable `ttrssd':
 #
 # ttrssd_enable="YES"
+#
+# Optional:
+#    ttrssd_run_as_user            The user to run as (default www)
 
 . /etc/rc.subr
 
@@ -18,6 +21,7 @@ rcvar=ttrssd_enable
 # read settings, set default values
 load_rc_config "${name}"
 : ${ttrssd_enable="NO"}
+: ${ttrssd_run_as_user="%%WWWOWN%%"}
 
 long_name="Tiny Tiny RSS updating feeds daemon."
 required_files="%%WWWDIR%%/config.php"
@@ -30,7 +34,7 @@ phpupd="%%WWWDIR%%/update_daemon2.php"
 ttrssd_log="/var/log/${name}.log"
 
 command="/usr/sbin/daemon"
-command_args="-rR 10 -H -u %%WWWOWN%% \
+command_args="-rR 10 -H -u $ttrssd_run_as_user \
 		-P $pidfile -p $cpidfile \
 		-o $ttrssd_log sh -c \
 		'$initdb_php --update-schema=force-yes; \


metalefty:
Ping to maintainer @derekschrock

rwv37:
Thanks @metalefty - I will submit a report on Bugzilla. A question, though, if you don't mind: Do I need to literally submit a patch there (i.e. a patch file), or just a message alerting the maintainer to the PR here on GitHub?

rwv37:
(I mean for the other changes in the future - I understand that you've already alerted the maintainer for this one)

metalefty:
Either is fine. The latter option would be fine in case when you've already raised a pull request. You don't need to attach a patch; just include a link to this PR instead. 

Attaching a patch created by `git format-patch` is also a good method when you don't raise a pull request on GitHub. Don't forget to include the port origin such as `www/tt-rss` in the Bugzilla summary (title), it pulls the trigger of automated notification.

derekschrock:
The init. pass of this seems alright however `www/tt-rss/update.php` and `www/tt-rss/update_daemon2.php` permissions need to be updated to allow execution by other.  If this is changed post pkg-install it would be reverted on update.

https://github.com/freebsd/freebsd-ports/blob/main/www/tt-rss/pkg-plist#L4

```
...
@mode ugo=rx
...
```

I believe it's up to the user to make sure `www/tt-rss/config.php` permissions are set correctly.  This is `0400` by default.

rwv37:
Thanks, @derekschrock. The way I did it on my own machine was just to manually ``chown`` everything that had been owned by ``www`` to my desired user instead, post-install, but based on your comment I'm guessing that sort of thing might get reverted on update too? I'll cancel this PR and when I get a chance (probably today or tomorrow, but not 100% certain), I'll make the update you suggested and resubmit a different PR.

By the way, I'm new to using GitHub too (except for commenting on stuff and little things like that), so if I'm supposed to do something other than "cancel this one and resubmit a new one", please let me know. Thanks again.

PR #306, opened 2024-10-03
allow ttrssd to run as a specified user (take two)

rwv37:
The Tiny Tiny RSS update daemon, ``ttrssd``, from package ``www/tt-rss``, currently runs as user ``www``. This change is meant to enable (optionally) running as a different user, by setting a new flag ``ttrssd_run_as_user`` in ``/etc/rc.conf``. For example, to run as the user ``mmm``, add the following line to ``/etc/rc.conf``:

```
ttrssd_run_as_user="mmm"
```

If this new flag is not specified, it defaults to running as ``www`` (reproducing the behavior from before this change).

I've modified [my previous PR](https://github.com/freebsd/freebsd-ports/pull/304) for this as suggested by @derekschrock:

### DIFF
--- a/www/tt-rss/files/ttrssd.in
+++ b/www/tt-rss/files/ttrssd.in
@@ -7,6 +7,9 @@
 # Add the following lines to /etc/rc.conf to enable `ttrssd':
 #
 # ttrssd_enable="YES"
+#
+# Optional:
+#    ttrssd_run_as_user            The user to run as (default www)
 
 . /etc/rc.subr
 
@@ -18,6 +21,7 @@ rcvar=ttrssd_enable
 # read settings, set default values
 load_rc_config "${name}"
 : ${ttrssd_enable="NO"}
+: ${ttrssd_run_as_user="%%WWWOWN%%"}
 
 long_name="Tiny Tiny RSS updating feeds daemon."
 required_files="%%WWWDIR%%/config.php"
@@ -30,7 +34,7 @@ phpupd="%%WWWDIR%%/update_daemon2.php"
 ttrssd_log="/var/log/${name}.log"
 
 command="/usr/sbin/daemon"
-command_args="-rR 10 -H -u %%WWWOWN%% \
+command_args="-rR 10 -H -u $ttrssd_run_as_user \
 		-P $pidfile -p $cpidfile \
 		-o $ttrssd_log sh -c \
 		'$initdb_php --update-schema=force-yes; \
--- a/www/tt-rss/pkg-plist
+++ b/www/tt-rss/pkg-plist
@@ -1,7 +1,7 @@
 %%DATADIR%%/httpd-tt-rss.conf
 @sample %%EXAMPLESDIR%%/newsyslog.sample etc/newsyslog.conf.d/ttrssd.conf
 @group %%WWWGRP%%
-@mode ug=rx
+@mode ugo=rx
 %%WWWDIR%%/update.php
 %%WWWDIR%%/update_daemon2.php
 @group


0mp:
Hey! Thanks for the patch!

Have you tried setting `ttrssd_user=mmm` in rc.conf? This is the canonical way of running a daemon as a specific user.

derekschrock:
I think I looked at that a while ago but thought it wouldn't work (or work as well) since daemon would be running as non-root and it was easier to to have root own all the logs, pids, etc...

Looking at other things in www that run a service (don't know how I missed them before) but they pre command install /dev/null with permissions/ownership for things like pids.  However, that would leave logs to either /var/log to it own dir /var/log/tt-rss/...  If so we would need add something to UIDs and/or GIDs?  Otherwise continue with the default %%WWWOWN%% (www)?

rwv37:
@0mp, I believe I did try that -- I think it was the first thing I tried -- but I can't say for sure at the moment, and unfortunately I don't have time to fall back from what I've got working and give it a try. I will try when I get a chance. But in the meantime, I can more confidently say a few things that may be related:

I took a look at my existing ``/etc/rc.conf/`` and saw that I had previously explicitly set up ``caddy`` to run as a certain user, via ``caddy_user``. I then looked at ``/usr/local/etc/rc.d/caddy`` to check out how it is set and used there, and tried to set up ``/usr/local/etc/rc.d/ttrssd`` in a similar way, adding ``ttrssd_user`` to it and using that to replace its hardcoded ``www``.

That resulted in some sort of error that, to me, seemed very wacky. Unfortunately, I forget what it was, but in trying to figure it out, I was led to the function ``rc_run_command`` (in ``/etc/rc.subr``). In there, I noticed something that seemed to indicate that variables of the form ``(pkg)_user`` are somehow treated specially. So, I tried just changing my ``ttrssd_user`` to ``ttrssd_run_as_user``, and everything started working just fine.

0mp:
The current implementation of ttrssd does not support the ttrssd_user variable. I'll see if I can change that.

@rwv37 what's your use case for running this daemon as a user other than the standard www?

0mp:
For the record, 

```sh
service ttrssd enable
sysrc ttrssd_user=nobody
service ttrssd start
```

results in

```console
Starting ttrssd.
Illegal option --
<pre>Exception while creating PDO object:could not find driver</pre>eval: ": not found
/usr/local/etc/rc.d/ttrssd: WARNING: failed to start ttrssd
```

rwv37:
Thanks @0mp.

Regarding my use case: Nothing specific, really; just paranoia, I guess. A lot of things are configured (by default) as ``www``, many of which do not actually need access to some or all of the others. As a general rule, I'd rather things not have permissions/abilities that they don't actually need.

And yes, that looks like the "wacky" error I was getting that eventually led me to replace the ``ttrssd_user`` that I had added with ``ttrssd_run_as_user`` instead.

0mp:
If security is the use case, then I'd suggest that introducing a ttrssd user to UIDs in the ports tree and then using that user/group for the service script is the way to go. This way everyone benefits from having extra security and there is no need to introduce a non-standard variable to the service script. 

The second best approach is to modify the ttrssd service script to handle ttrssd_user correctly. Currently, the problem is probably that all the directories belong to www and so if you just change ttrssd_user then the service hits some permission problems. Nevertheless, ttrssd_user should be the variable we use, not ttrssd_run_as_user.

rwv37:
OK, thanks @0mp. I think that given your explanation and my schedule, I have too little time in the foreseeable future to learn everything I'd need to in order to do this right. So, I'll be closing this pull request.

Thanks again everybody!

0mp:
What comes to my mind is that you may consider jailing ttrssd with the new svcj mechanism.

0mp:
I opened a PR on Bugzilla: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=282245

PR #310, opened 2024-10-12
Fix man page reference for paperless-ngx

maltekrupa:
Hi,

the config file for paperless-ngx has a reference to a man page which does not exist in the pkg.

Best,
Malte


arrowd:
Merged, thanks.

PR #311, opened 2024-10-16
Update step-ca.in to correct networking requirement

rwv37:
The rcorder command would report:

"rcorder: requirement `networking' in file `/usr/local/etc/rc.d/step-ca' has no providers."

This commit fixes the issue by changing the REQUIRE of "networking" to be "NETWORKING" instead.


rwv37:
Bugzilla: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=282139

rwv37:
Makefile distversion and portrevision updated to reflect changes that have been made there since this pull request was opened.

arrowd:
Merged, thank you.

PR #313, opened 2024-10-17
Update Makefile - require ripgrep

rwv37:
textproc/ripgrep-all will exit with a failure message when invoked if ripgrep is not also installed. This commit changes the Makefile so that textproc/ripgrep is listed as a runtime dependency, and therefore will be automatically built and installed along with textproc/ripgrep-all.


rwv37:
FreeBSD Bugzilla bug report is [here](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=282151)

clausecker:
Will commit the patch on Bugzilla soon.

PR #315, opened 2024-10-25
misc/mc: fix packaging WITHOUT_EDITOR

smyru:
If the operator disables building the port with editor, packaging of the port fails, since pkg-plist will believe editor's man shall be installed anyway.


arrowd:
Pushed, thank you.

PR #325, opened 2024-11-04
security/osv-scanner: Update to 1.9.1

khorben:
This now requires Go 1.23.

Changelog:	https://github.com/google/osv-scanner/releases/tag/v1.9.1
Sponsored by:   The FreeBSD Foundation


khorben:
Moved to https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=282822.

PR #332, opened 2024-12-15
Updates for some of Gnome stuff

arrowd:
GVFS was compile-tested against all its consumers. json-glib was compile-tested selectively.


arrowd:
@vishwin You might be interested ^

diizzyy:
Please migrate to DISTVERSION whenever possible for consistency

devel/libgdata probably doesn't need a PORTREVISION bump as it only seems to affect building?

Should lib/libmsgraph-0.so.1 really be so.1 and not so.0 ?

arrowd:
> Should lib/libmsgraph-0.so.1 really be so.1 and not so.0 ?

This question should be directed to the upstream.

vishwin:
The latest uhttpmock has actually been ported to libsoup3, I will update it by temporarily moving the current version to a separate port, so you will need to rebase afterward.

vishwin:
I committed the uhttpmock stuff, libgdata can be removed from this sequence. There's no indication they will move to libsoup3 anytime soon so the existing uhttpmock is moved to a separate port.

filesystems/gvfs/Makefile
```diff
@@ -15,7 +14,8 @@ LICENSE_FILE=	${WRKSRC}/COPYING
 BUILD_DEPENDS=	docbook-xsl>=0:textproc/docbook-xsl
 LIB_DEPENDS=	libgcrypt.so:security/libgcrypt \
 		libdbus-1.so:devel/dbus \
-		libgcr-base-3.so:security/gcr3 \
+		libgcr-4.so:security/gcr \
+		libmsgraph-0.so:net/msgraph \
```
vishwin, line 18: this should be placed with `GOA` or a separate option that implies/depends on `GOA`

arrowd:
Comments addressed.

net/msgraph/Makefile
```diff
@@ -0,0 +1,33 @@
...
+USE_GNOME=	glib20 introspection:build
+USE_LDCONFIG=	yes
+
+DOCSDIR=	${PREFIX}/share/doc/${PORTNAME}-0
+PORTDOCS=	*
+
+OPTIONS_DEFINE=		DOCS TEST
+
+DOCS_MESON_TRUE=	gtk_doc
+DOCS_BUILD_DEPENDS=	gtkdoc-scan:textproc/gtk-doc \
```
vishwin, line 27: Despite the option name, only gi-docgen gets called looks like

vishwin:
ship it, after accounting for msgraph

PR #333, opened 2024-12-16
Fix glusterfs

darkain:
Revert the "broken" status for GlusterFS and add an extra LDFLAGS line that allows the port to compile and run as expected.

I'm actively using this and have been using it for a few weeks now on FreeBSD 14.1.

I'll also start looking into updating the port to a more recent upstream version of GlusterFS.


fpedrei:
Thanks for looking into this, please don't forget to increase PORTREVISION.

Do you think it is possible to increase glusterfs version (to 10 or 11)?

PR #335, opened 2024-12-23
misc/shared-mime-info: add new mimesubdirs 

Awoonyaa:
@arrowd


arrowd:
Pushed with few adjustments, thank you.

PR #336, opened 2024-12-28
Update Makefile

linimon:
converters/p5-Digest-SHA has been listed in Requirements.pm for many years.  Add it.

(I do not know why Bugzilla works without it.)

While here, clarify which Bugzilla (ours) the PRs are part of.


arrowd:
Pushed, thank you.

I had to change the dependency to `seecurity/p5-Digest-SHA`, because `converters/p5-Digest-SHA` does not exist.

linimon:
> I had to change the dependency to security/p5-Digest-SHA, because converters/p5-Digest-SHA does not exist.

oops, thanks.

mcl

PR #337, opened 2025-01-04
Slim down devel/llvm11

brooksdavis:
This set of patches removes consumers of devel/llvm11 where possible (most didn't actually use it), removes if from USES=compiler and USES=llvm, and pares the default build down to just building the core LLVM bits. Unfortunately, it's used to bootstrap old ghc versions on aarch64 and armv7 so we're stuck with it for now, but we can cut its size dramatically.


arrowd (approved):

jbeich:
Did you forget to remove `llvm11` flavor in `intel-graphics-compiler` + `intel-compute-runtime`? Only llvm11 is "production" for the current version (in ports, not upstream). See also [bug 281566](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=281566)

jbeich:
Doh, forgot intel-graphics-compiler@llvm11 also uses vc-intrinsics@llvm11, not just spirv-llvm-translator@llvm11 and opencl-clang@llvm11.

brooksdavis:
> Did you forget to remove `llvm11` flavor in `vc-intrinsics` + `intel-graphics-compiler` + `intel-compute-runtime`? Only llvm11 is "production" for the current version (in ports, not upstream). See also [bug 281566](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=281566)

I missed them because my grep for llvm.*11 missed `${14 13 12 11:L:S/^/llvm/}` (that idiom saved 4 characters and destroyed grepability, seems like a poor tradeoff)

PR #346, opened 2025-02-13
Update dotnet/lang to 9.0.2

sec:
Tested build for both 13 and 4 under x64/arm64 hosts.
Updated bootstrap SDK used to 9.0.100
Adjust pkg-plist to make it more flexible for future releases.


arrowd:
`lang/dotnet-runtime` should be update too, I suspect?

sec:
@arrowd updated.

arrowd:
Landed, thank you.

PR #359, opened 2025-03-07
sysutils/framework-system: Update to v0.2.1

JohnAZoidberg:
Not the regular release, had to make some changes. FreeBSD has older rust dependencies and cannot ignore packages that aren't actually used during build.

Adds support for Framework's new platforms in 2025, among other changes.


JohnAZoidberg:
Can I avoid making changes in my project for ports?
If I just build directly with cargo it works, but with the ports tree I need to remove unused dependencies and downgrade a lot of packages.

JohnAZoidberg:
> Can I avoid making changes in my project for ports?

I know I can apply patches from the port here.
But it would be better to fix in the port config instead of patching the source awkwardly.

lwhsu:
> Can I avoid making changes in my project for ports? If I just build directly with cargo it works, but with the ports tree I need to remove unused dependencies and downgrade a lot of packages.

Sorry that I don't quite understand the question, do you mean the rust dependency in Makefile.crates?

If that's the case, it's generated via `make cargo-crates` if there is `USES= cargo` in Makefile.

sysutils/framework-system/Makefile
```diff
@@ -1,5 +1,5 @@
 PORTNAME=	framework-system
-DISTVERSION=	0.1.0
+DISTVERSION=	0.2.1
 PORTREVISION=	6
```
lwhsu, line 3: PORTREVISION should be removed (implies 0) when DISTVERSION is updated.
The newer package version will be `0.2.1` and no need to be `0.2.1_6` as it's already greater than `0.1.0_6`

JohnAZoidberg:
> Can I avoid making changes in my project for ports?

I know I can apply patches from the port here.
But it would be better to fix in the port config instead of patching the source awkwardly.



> > Can I avoid making changes in my project for ports? If I just build directly with cargo it works, but with the ports tree I need to remove unused dependencies and downgrade a lot of packages.
> 
> Sorry that I don't quite understand the question, do you mean the rust dependency in Makefile.crates?
> 
> If that's the case, it's generated via `make cargo-crates` if there is `USES= cargo` in Makefile.

See the changes I had to make here
https://github.com/FrameworkComputer/framework-system/commits/freebsd-v0.2.1/

arrowd:
Why did you remove these lines in the first place? What error were you getting?

sysutils/framework-system/Makefile
JohnAZoidberg: Oh ok thanks, will do

JohnAZoidberg:
It didn't find the packages, and the other packages it didn't find the right version

arrowd:
Like @lwhsu correctly noticed, you didn't update Makefile.crates. This file defines versions of the rust dependencies that will be used to build the port. This list can be generated automatically with `make cargo-crates`. You'll also need to run `make makesum` after that to refresh `distinfo`.

JohnAZoidberg:
Ok, you're right, i ran `make cargo-crates` but didn't pipe it to `Makefile.crates`.

But now it can't build anymore, could you please try if this can build for you?
Now it can't  find smbios-lib, redox_hwio and uefi.
But those haven't changed, so they should still work.

JohnAZoidberg:
Is there any CI to help check if this can build or not?

arrowd:
We don't have a CI yet, but you can replicate what the official FreeBSD builder does by setting up Poudriere and running `poudriere testport -j yourjail -p yourports sysutils/framework-system`.

arrowd:
The port builds fine for me, actually.

arrowd:
Let me know if I should land this.

JohnAZoidberg:
> The port builds fine for me, actually.

with poudriere? Then yes please, feel free to merge it :D
Maybe try to run the binary and see if it can start at least. You could check `framework_tool --version`

JohnAZoidberg:
Thanks for helping out!

lwhsu:
Merged as 5bcfc0c0a7081bb694b592971b313972576c17f7

PR #360, opened 2025-03-08
iio-sensor-proxy: init at 3.6

JohnAZoidberg:
Preparation for Framework 12 (Convertible) g-sensor to let the screen rotate.

monitor-sensor can run, but we don't have any kernel interfaces for sensors in FreeBSD yet.


JohnAZoidberg:
Draft because it's not useful yet, but I wouldn't mind if it got merged already.
We can enable kernel interfaces and upgrade this package later.

lwhsu:
> Draft because it's not useful yet, but I wouldn't mind if it got merged already. We can enable kernel interfaces and upgrade this package later.

Yeah, I think it's fine to have things prepared. Or at least letting it ready and people can test with the kernel changes.

I've done some adjustments to the Makefile to follow the ports' standard:

https://gist.github.com/lwhsu/fd9f05f055db2f13765fa2fea7e1a371

Also, it seems not buildable in the poudriere for now:

https://gist.github.com/lwhsu/759f4f824169525cb46640d951386843

comms/iio-sensor-proxy/Makefile
```diff
@@ -0,0 +1,25 @@
...
+USES=		pkgconfig meson
+USE_GNOME+=	glib20
+USE_GITLAB=	yes
+GL_SITE=	https://gitlab.freedesktop.org
+GL_ACCOUNT=	hadess
+
+LICENSE=	GPLv3
+LICENSE_FILE=	${WRKSRC}/COPYING
+
+LIB_DEPENDS+=	libudev.so:devel/libudev-devd \
```
JohnAZoidberg, line 18:
```suggestion
BUILD_DEPENDS=	evdev-proto>0:devel/evdev-proto
LIB_DEPENDS+=	libudev.so:devel/libudev-devd \
```

@lwhsu I think this should fix the poudriere build

JohnAZoidberg:
> I've done some adjustments to the Makefile to follow the ports' standard:

Thanks, I applied your formatting.

lwhsu:
I got this in poudriere test build:

```
====> Checking for pkg-plist issues (check-plist)
===> Parsing plist
===> Checking for items in STAGEDIR missing from pkg-plist
Error: Orphaned: disabled/80-iio-sensor-proxy.rules
Error: Orphaned: disabled/iio-sensor-proxy.service
Error: Orphaned: libexec/iio-sensor-proxy
===> Checking for items in pkg-plist which are not in STAGEDIR
Error: Missing: usr/lib/iio-sensor-proxy
Error: Missing: usr/lib/udev/rules.d/80-iio-sensor-proxy.rules
===> Error: Plist issues found.
```

The first part means that this port installs these files but we don't have them listed in pkg-plist, I'm not sure if those files under `disabled/` is really needed, though.

The second part mean that we have them in pkg-plist but not installed, it seems `iio-sensor-proxy` is actually be in `libexec/` (and that looks reasonable)

What I'm not sure is that should we do something to teach it install `80-iio-sensor-proxy.rules` to the right place we want?

comms/iio-sensor-proxy/Makefile
```diff
@@ -0,0 +1,27 @@
...
+
+LICENSE=	GPLv3
+LICENSE_FILE=	${WRKSRC}/COPYING
+
+BUILD_DEPENDS=	evdev-proto>0:devel/evdev-proto
+LIB_DEPENDS+=	libudev.so:devel/libudev-devd \
+		libgudev-1.0.so:devel/libgudev \
+		libpolkit-gobject-1.so:sysutils/polkit
+
+USES=		pkgconfig meson
```
arrowd, line 17: USES values should be sorted.

comms/iio-sensor-proxy/pkg-plist
```diff
@@ -0,0 +1,5 @@
+bin/monitor-sensor
+usr/lib/iio-sensor-proxy
```
arrowd, line 2: This looks suspicious. It feels like we should have `MESON_ARGS= -Dlibexecdir=${PREFIX}/libexec` in the Makefile to make this file go into a correct dir.

comms/iio-sensor-proxy/pkg-plist
```diff
@@ -0,0 +1,5 @@
...
+usr/lib/udev/rules.d/80-iio-sensor-proxy.rules
```
arrowd, line 3: If this file is still installed despite the `-Dudevrulesdir=disabled` arg, then we need to prepend `@comment ` there to prevent it from being packaged into the final package.

arrowd (changes requested):

PR #361, opened 2025-03-14
Linux rl9 stable diffusion demo

Voxnod:
@arrowd


arrowd:
Landed.

cederom:
Hello @Voxnod this port seems to stopped working after switch of linuxlator base from C7 to RL9. Can you please take a look? :-)

This is also kind of related to https://github.com/shkhln/libc6-shim/issues/14 where RTX5070 does not seem to work with libc6-shim and we are looking for some common ready to use AI/ML ports to test out stuff :-)

We could not find you on FreeBSD Bugzilla, below is the log of problems (Python 3.9 modules) and additional patch from TA that enabled building port with nvidia*-devel packages :-)

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=291139

Thanks :-)

PR #367, opened 2025-04-04
net/miniupnpd: Update to 2.3.9

Self-Hosting-Group:
- Update miniupnpd to 2.3.9
- Build daemon with `--disable-pppconn` to remove legacy workaround no longer included in other implementations since >15y
- Remove unnecessary OpenSSL dependency
- Remove obsolete files


arrowd:
I'm afraid you'd need to submit this via Bugzilla, because this port has `MAINTAINER` set, so we'll need maintainer's approval to merge this in.

bsdimp:
As an experiment, I've created https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=286295 to loop in the maintainer. This is a manual process, so ping me if things appear stalled. We'll see how it works.

bsdimp:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=278576

bsdimp:
I just pointed the bug here. I didn't redact anything.

The other bug might have done those things, but I've not looked. Once that lands, we can iterate if need be.

arrowd:
Merged.

arrowd:
Thanks for the heads-up, I pulled that patch in.

PR #369, opened 2025-04-08
Update Dub and fix D ports

arrowd:
Tested with `poudriere bulk -t` on all Dub consumers.

@alonsobsd please approve


alonsobsd:
Approved. Thanks!

PR #371, opened 2025-04-08
Update x11-themes/qgnomeplatform

h-2:
* the first commit brings the port to the most recent version
* the second commit renames the port to `-qt5`
* I will create a separate PR that adds `x11-themes/qgnomeplatform-qt6`


rosorio:
I'm sorry but upstream repo seems discontinued since two years ago, and the port has no maintainer.
This looks more a candidate for deprecation than a fresh update.

h-2:
Well, it's the only way to get consistent theming between qt ang gtk 🤷🏻 

I am assuming that the toolkits will be around for a while?

PR #380, opened 2025-04-21
[NEW PORT] net-mgmt/prometheus-bird-exporter

llfw:
Prometheus metric exporter for the BIRD routing daemon.

### DIFF
--- a/net-mgmt/prometheus-bird-exporter/Makefile
+++ b/net-mgmt/prometheus-bird-exporter/Makefile
@@ -1,0 +1,26 @@
+PORTNAME=	prometheus-bird-exporter
+DISTVERSION=	1.4.3
+CATEGORIES=	net-mgmt
+
+MAINTAINER=	ivy@FreeBSD.org
+COMMENT=	Prometheus state exporter for BIRD routing daemon
+WWW=		https://github.com/czerwonk/bird_exporter
+
+LICENSE=	MIT
+LICENSE_FILE=	${WRKSRC}/LICENSE
+
+USES=		go:modules
+
+USE_GITHUB=	yes
+GH_ACCOUNT=	czerwonk
+GH_PROJECT=	bird_exporter
+
+USE_RC_SUBR=	bird_exporter
+
+GO_MODULE=	github.com/czerwonk/bird_exporter
+GO_MOD_DIST=	https://raw.githubusercontent.com/czerwonk/bird_exporter/refs/tags/${DISTVERSION}/
+GO_TARGET=	.:bird_exporter
+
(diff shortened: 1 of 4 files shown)


net-mgmt/prometheus-bird-exporter/Makefile
```diff
@@ -0,0 +1,27 @@
...
+
+LICENSE=	MIT
+LICENSE_FILE=	${WRKSRC}/LICENSE
+
+USES=		go:modules
+
+USE_GITHUB=	yes
+GH_ACCOUNT=	czerwonk
+GH_PROJECT=	bird_exporter
+GH_TAGNAME=	${DISTVERSION}
```
arrowd, line 17: `GH_TAGNAME` defaults to `DISTVERSION`.

net-mgmt/prometheus-bird-exporter/Makefile
```diff
@@ -0,0 +1,27 @@
...
+
+USE_RC_SUBR=	bird_exporter
```
arrowd, line 19: If there might be many bird exporters in the wild, then the rc script should be named `prometheus_bird_exporter`. This would also match the PORTNAME better.

arrowd (changes requested):

net-mgmt/prometheus-bird-exporter/Makefile
llfw: i named the rc.d script "bird_exporter" because this is what the daemon is called, but the package is called prometheus-bird-exporter for consistency in the ports tree.

this is similar to how databases/prometheus-postgresl-exporter is called postgres_exporter.

i'd rather not rename it since i doubt there will be another bird_exporter.

arrowd: Sure, whatever you feels is right. I have no idea about all this stuff.

net-mgmt/prometheus-bird-exporter/Makefile
llfw: fixed

arrowd:
LGTM.

arrowd:
The commit message should be reworded to something like
```
net-mgmt/prometheus-bird-exporter: Prometheus metric exporter for the BIRD routing daemon
```

llfw:
do we not include [NEW PORT] for new ports anymore?

arrowd:
Haven't seen it in ages. And the pre-commit hook requires the message to start with `category/port:` anyways.

llfw:
okay, amended.  although the new subject line is way over the 50 character limit...

kevans91 (approved):

arrowd (approved):

PR #382, opened 2025-04-23
databases/galera: update to 26.4.22

fauust:
First contribution here, sorry if something is missing.

### DIFF
--- a/databases/galera26/Makefile
+++ b/databases/galera26/Makefile
@@ -1,5 +1,5 @@
 PORTNAME=	galera
-PORTVERSION=	26.4.21
+PORTVERSION=	26.4.22
 PORTREVISION=	1
 DISTVERSIONPREFIX=	mariadb-
 CATEGORIES=	databases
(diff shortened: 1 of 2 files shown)


arrowd:
We need a maintainer approval for this and it can only be obtained from Bugzilla. Can you please resubmit this change to bugs.freebsd.org?

fauust:
Thanks @arrowd !
I did a request for login on bugs.freebsd.org but still no response, but that's probably normal.
If you have any pointers to documentation on how we are supposed to update FreeBSD ports I would appreciate it since I am a bit lost and I don't understand where upstream code is (here or on bugs.freebsd.org). Do I understand correctly that I need to feel a report somewhere in https://bugs.freebsd.org/bugzilla/enter_bug.cgi?component=Individual%20Port%28s%29&product=Ports%20%26%20Packages?

arrowd:
Yep, that's correct. And put `databases/galera:` in front of the SUmmary field to get the maintainer notified automatically.

PR #386, opened 2025-05-06
Uses/cargo: use --locked flag on install

ckoehne:
By default cargo install ignores the lock file [1]. This will cause our
build to fail easily. First of all, when building with tools like
poudriere, network access will be disabled in install stage. Fetching a
new crate won't be possible. Additionally, a newer crate version might
be incompatible with our crate or even more worse unbuildable at all.
For that reason, force cargo to honor the lock file on install.

[1] https://doc.rust-lang.org/cargo/commands/cargo-install.html#dealing-with-the-lockfile

Sponsored by:		Beckhoff Automation GmbH & Co. KG

### DIFF
--- a/Mk/Uses/cargo.mk
+++ b/Mk/Uses/cargo.mk
@@ -341,6 +341,7 @@ do-build:
 do-install:
 .    for path in ${CARGO_INSTALL_PATH}
 	@${CARGO_CARGO_RUN} install \
+		--locked \
 		--no-track \
 		--path "${path}" \
 		--root "${STAGEDIR}${PREFIX}" \


arrowd:
The litellm change would require an approval from @tagattie

ckoehne:
> I pushed an update to py-litellm a while ago though it seems unrelated to the proposed change to cargo.mk 😄

Yeah, sry for that noise. I guess the commit slipped in because the github mirror hasn't synced yet.

diizzyy:
Sounds reasonable to me but please create a PR in bugzilla

mat813:
Well, it might feel reasonable, but it has to be possible to disable that knob, crates gets pulled, and lock files become unusable.

ckoehne:
> Sounds reasonable to me but please create a PR in bugzilla

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=286761

PR #391, opened 2025-05-07
Fixes various issues in security/tpm2-abrmd

puffetto:
Issue 1: Name collision on function write_all():

This port in file tpm2-abrmd-3.0.0/src/util.h at line 82 declares a function "ssize_t write_all(GOutputStream *ostream, const uint8_t *buf, const size_t size);", implemented in work/tpm2-abrmd-3.0.0/src/util.c.

The tpm2-abrmd, at runtime, when trying to access directly the device /dev/tpm0, loads dynamically [using dlopen()] the library /usr/local/lib/libtss2-tcti-device.so.0.0.0 provided by port security/tpm2-tss,

The function write_all() is implemented by port security/tpm2-tss in file src/util/io.c at line 85 with prototype "ssize_t write_all(SOCKET fd, const uint8_t *buf, size_t size)".

Basically the two functions do the same thing, but the one in tpm2-abrmd takes as first parameter a GOutputStream while the one in tpm2-tss expects a file descriptor; at runtime the wrong function gets called and this causes a coredump.

As the function in tpm2-abrmd acts on a GOutputStream, the most sane approach seems to be prepending a g_ to its name, avoiding the name clash.

The issue is likely triggered by the fact that the library libtss2-tcti-device.so.0.0.0 is not loaded by the linker, but explicitly.

-----
Issue 2: tpm2-abrmd-devd.conf is useless and does not work

The port installs a tpm2-abrmd-devd.conf in /usr/local/etc/ aimed at fixing the ownership and permission of /dev/tpm0

This does not work: /dev/tpm0 (if available) is created by the kernel at an early boot stage, devd does not get any notification.

The proper place to fix the permissions and ownership of the device is /etc/devfs.conf, as it does not look like a great idea to programmatically change a file in /etc I added, as most ports to in these cases, a message to the sysadmin asking to do it.

-----
Issue 3: We do not have a group named "root"

The port installs a /usr/local/etc/dbus-1/system.d/tpm2-abrmd.conf a policy giving permissions to a group name "root".

Unix does not have a group named "root", gid 0 is named "wheel"; fixed.

-----
Issue 4: Inconsistent naming of /usr/local/etc/rc.d/tpm2-abrmd

The program name, unfortunately contains a dash (tpm2-abrmd), which is not sane in rc scripts; the provided rc script correctly uses the service name "tpm2_abrmd".

Still, the script filename is "tpm2-abrmd", while docs at https://docs.freebsd.org/en/articles/rc-scripting/ clearly state "The content of the name variable needs to match the script name".

I renamed the script tpm2_abrmd, as I could not rename the service to tpm2-abrmd ["the filename shall also not contain characters which may be troublesome in scripting (e.g., do not use a hyphen "-" and others)"].


security/tpm2-abrmd/files/patch-src_util.c
```diff
@@ -0,0 +1,30 @@
...
+-                 size - written_total);
++        g_debug ("%s: writing %zu bytes to ostream from %zu", __func__,
++                 size - written_total, buf);
+         written = g_output_stream_write (ostream,
+                                          (const gchar*)&buf [written_total],
+                                          size - written_total,
+                                          NULL,
+                                          &error);
+-        switch (written) {
++        g_debug ("Passed");
```
arrowd, line 26: Debugging leftover?

security/tpm2-abrmd/Makefile
```diff
@@ -17,7 +17,7 @@ RUN_DEPENDS=	dbus-daemon:devel/dbus
 USES=		gmake libtool pkgconfig gnome
 USE_LDCONFIG=	yes
 USE_GNOME=	glib20
-USE_RC_SUBR=	tpm2-abrmd
+USE_RC_SUBR=	tpm2_abrmd
```
arrowd, line 20: This requires renaming `files/tpm2-abrmd.in` accordingly.

arrowd (changes requested):

security/tpm2-abrmd/Makefile
puffetto: Harmless, but reolved in 12cb3d0fd6de

security/tpm2-abrmd/files/patch-src_util.c
puffetto: Yes, harmless, but fixed in 12cb3d0fd6de

arrowd (approved):

tstruk (approved):

arrowd:
Merged, thanks everyone.

PR #392, opened 2025-05-07
Fixes issue in security/tpm2-tss trying to access /dev/tpmrm0

puffetto:
FreeBSD does not have an in-kernel Resource Manager for TPM, the tools must access the /dev/tpm0 device, either directly or through an userland RM like tpm2-abrmd.

The default behavior of the original code is to try the in kernel RM on /dev/tpmrm0 as first choice, this causes warnings and errors and forces the user to override the behavior with options or environment variables.

This patch removes the attempt to use /dev/tpmrm0 so that the programs "just work" using the tpm0 device or through tpm2-abrmd.


puffetto:
Commit 2113074ccbc5e3f1ae76c6d310f7bff4a73bed33 fixed a mistake I did.
Rule #315; Never edit the diff of a diff by hand... or at least test it! 🤣

tstruk (approved):

arrowd:
Merged, thanks everyone.

PR #394, opened 2025-05-11
Fix Redis 8.0 license

LiorKogan:


ig-raiderler:
Impressive project structure! The architecture really demonstrates solid design principles.

yasuhirokimura:
`LICENSE_COMB` is defined on line 75 of `Mk/bsd.licence.db` as following.

```
# LICENSE_COMB		- Set to "dual" or "multi", for OR/AND behavior
```

And upstream says as following in release notes of Redis 8.0.0.

```
License change: licensed under your choice of

(a) the Redis Source Available License 2.0 (RSALv2); or
(b) the Server Side Public License v1 (SSPLv1); or
(c) the GNU Affero General Public License (AGPLv3)
```

So `LICENSE_COMB` should be set to `dual` in this case.

And `AGPLv3` is already defined in `Mk/bsd.licenses.db.mk`. So we need not to set `LICENSE_NAME_AGPLv3`and `LICENSE_PERMS_AGPLv3` explicitly in 'Makefile`.

PR #397, opened 2025-05-18
sysutils/sysinfo: fix default config

pbruenn:
Commit 80813f1400207f7e8c90b918ade272328eb4caeb broke the default installation:
  $ sysinfo -a
  sysinfo: ERROR: Configuration file was found

Since, there was no explanation in the commit message we assume that was an oversight. So here we go and restore the old behaviour.

Fixes: 80813f1400207f7e8c90b918ade272328eb4caeb

cc: @fernape

### DIFF
--- a/sysutils/sysinfo/Makefile
+++ b/sysutils/sysinfo/Makefile
@@ -1,5 +1,6 @@
 PORTNAME=	sysinfo
 PORTVERSION=	1.0.4
+PORTREVISION=	1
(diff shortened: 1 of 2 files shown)


fernape:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=286909

fernape:
Fixed,

Thanks for the heads up!

PR #398, opened 2025-05-18
sysutils/podman: Use full path for alpine image

ricardobranco777:
Use full path for alpine image so newcomers to podman on FreeBSD aren't stopped by this error message:

`Error: short-name "alpine" did not resolve to an alias and no unqualified-search registries are defined in "/usr/local/etc/containers/registries.conf"`


ricardobranco777:
cc @DFR

PR #400, opened 2025-05-25
rpi cm4: enable USB(keyboard etc.) on boot

klaus4:
-- also patch in src required : https://github.com/freebsd/freebsd-src/pull/1711 

U-Boot 2025.04 (May 20 2025 - 05:56:19 +0200)

DRAM:  948 MiB (effective 7.9 GiB)
RPI Compute Module 4 (0xd03140)
Core:  211 devices, 16 uclasses, devicetree: board
MMC:   mmc@7e300000: 3, mmc@7e340000: 0
L.......

---BEFORE this patch:---
starting USB...
No USB controllers found.......

-- AFTER this patch:---
starting USB...
Bus usb@7e980000: USB DWC2
scanning bus usb@7e980000 for devices... 5 USB Device(s) found
       scanning usb for storage devices... 0 Storage Device(s) found
Hit any key to stop autoboot:  0 
U-Boot> usb info 3
config for device 3
4: Human Interface,  USB Revision 1.10
 - Mitsumi Electric Apple Extended USB Keyboard 
.....

### DIFF
--- a/sysutils/rpi-firmware/files/config_arm64.txt
+++ b/sysutils/rpi-firmware/files/config_arm64.txt
@@ -9,3 +9,9 @@ kernel=u-boot.bin
 [pi4]
 hdmi_safe=1
 armstub=armstub8-gic.bin
+
+[cm4]
+otg_mode=1
+
+[cm5]
+dtoverlay=dwc2,dr_mode=host


klaus4:
in src : https://github.com/freebsd/freebsd-src/pull/1711

markmi:
FYI: The default config.txt for:
```
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
NAME="Debian GNU/Linux"
VERSION_ID="12"
VERSION="12 (bookworm)"
VERSION_CODENAME=bookworm
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```
has, in part:
```
[cm4]
# Enable host mode on the 2711 built-in XHCI USB controller.
# This line should be removed if the legacy DWC2 controller is required
# (e.g. for USB device mode) or if USB support is not required.
otg_mode=1

[cm5]
dtoverlay=dwc2,dr_mode=host

[all]
```

klaus4:
> otg_mode=1

Okay, I'll test later today if OTG on the XHCI controller doesn't cause any problems, then we could change that.

klaus4:
> [cm4]

> otg_mode=1
> 
> [cm5]
> dtoverlay=dwc2,dr_mode=host

changed as you have suggested -
additionally I've discussed that with another cm4/cm5 owner

evadot:
You also need to bump PORTREVISION in the ports, after that we're all good.

klaus4:
> You also need to bump PORTREVISION in the ports, after that we're all good.

thanks,
you mean I should add PORTREVISION=   1  now or will you do that after/while upgrading ?(since I don`t see that variable in Makefile)

emaste:
> you mean I should add PORTREVISION= 1 now

Yes. This is needed so that the pkg has a new version and will get updated. You can make the change in your tree and re-push to update this pull request.

klaus4:
I see... I've held off on this patch for now, as it was only intended as a starting point for what I believe is a much-needed, larger RPI patch series. We have a few bug reports on the Pi4 mailing list: missing functionality for the cm4/5, faulty u-boot versions in releases, and so on. It needs some RPI-dedicated port/src-maintenance, possibly with patches for each individual board version... I know roughly what needs to be done and hope to find the time for it soon.

bsdimp:
what's the status of this additional work?

klaus4:
status= „forgotten on the attic“, or something like that .. 
but still relevant…
I’ve contacted a former Discord user who collaborated with me on creating a patched u-boot-version for the CM4 back then—one that detects not just the keyboard but also the NVMe SSD at boot. I’m checking to see if he’s still active; otherwise, if there’s interest, I’ll submit a port myself. I’ll have to get back up to speed on it, though.

PR #404, opened 2025-06-03
Update netbird.in

KeenanFalcon:
added several rcvar with default values to allow control from rc.conf.d, and pre_cmd to check and destroy orphaned interfaces.

The variable "netbird_env" is removed, as "service netbird status" gives an error that it's not found.


rosorio:
Looks like this pull request was [overcome by events](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=287762).

PR #407, opened 2025-06-09
Fix X11 ports to use GL_DEFAULT for mesa building

kenrap:
After adding graphics/mesa-devel to my `DEFAULT_VERSIONS` in my make.conf, I was continuously having x11-wm/plasma6-kwin failing to build using poudriere at the lib-depends phase with the following error:
```
pkg-static: mesa-devel-25.1.b.2196 conflicts with mesa-libs-24.1.7_1 (installs files into the same place).  Problematic file: /usr/local/include/EGL/eglext_angle.h
```
In addition, every time I try to rebuild plasma6-kwin, it would constantly force delete the existing built package for x11-servers/xwayland and rebuild it again. Lather-rinse-repeat.

The problem was that xwayland was still using mesa-dri as the build dep, not allowing the flexibility to use mesa-devel and going into a rebuild loop. This fix solves it.

After continuing to finish rebuilding my package cache, I was noticing similar "rebuild loop" patterns with xorg and xorg-server. So I fixed them the same way. It all builds and I'm now daily driving it.

### DIFF
--- a/x11-servers/xorg-server/Makefile
+++ b/x11-servers/xorg-server/Makefile
@@ -38,10 +38,10 @@ PLIST_FILES=	bin/${BINARY_NAME} \
 
 DESCR=	${.CURDIR}/pkg-descr-${FLAVOR}
 
-BUILD_DEPENDS+=	${LOCALBASE}/libdata/pkgconfig/dri.pc:graphics/mesa-dri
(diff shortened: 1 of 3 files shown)


evadot:
LGTM

arrowd:
Wait, this doesn't work. `GL_DEFAULT` evaluates to `mesa-libs`, not `mesa-dri`.

You should probably put `GL_DEFAULT` into mesa-dri port.

kenrap:
Closing this since the `COINST` build option in `mesa-devel` solves this problem.

PR #409, opened 2025-06-11
archivers/unarr: enable riscv64 build

mengzhuo:
LLVM now can build with 'd' ABI on riscv64 target


arrowd:
@jbeich You might be interested.

jbeich:
@arrowd, I don't maintain this port since 4bca597469758, and such trivial changes are covered by `portmgr blanket` (like cef8da8f0031 originally did).

@mengzhuo, "now" is vague and may not cover slightly outdated LLVM in base (e.g., -CURRENT hasn't updated to LLVM 20.* series yet). I don't have riscv64 hardware, so assuming you've tested on a recent FreeBSD version.

PR #410, opened 2025-06-16
devel/qt6-base: Pull in patches from upstream to fix a busy-loop bug

arrowd:
PR:		252934

@BSDKaffee OK to push?


BSDKaffee:
Looks good to me.

PR #413, opened 2025-06-27
chromium ports: Remove useless patch

arrowd:
On FreeBSD the 'environ' symbol is a part of CSU and not libc. This is a common source of linking errors in various ports that are accessing this variable. Luckily, the problem can only be observed if the code that is using "environ" ends up with a shared library that is linked to with -Wl,--no-undefined or -Wl,--no-allow-shlib-undefined.

This is not the case for Chromium, so this patch is safe to drop.

More info on the issue: https://reviews.freebsd.org/D30842


arrowd:
@rnagy Please take a look.

arrowd:
@rnagy Ping. I build-tested this change, FWIW.

arrowd:
@rnagy Ping again.

arrowd:
The same patch is still present in `www/iridium` and `www/ungoogled-chromium`.

rnagy:
Yes those will be removed when the ports get the update to their latest version.

arrowd:
All hunks merged.

PR #416, opened 2025-06-30
devel/kf6-solid: Add patch to revert showing ZFS datasets as devices

kenrap:
@arrowd


arrowd:
Merged, thanks a lot.

PR #417, opened 2025-07-01
net/mosquitto: don't set cafile in example config

sdirkwinkel:
Recently mosquitto started checking for invalid combinations of cafile/capath/certfile/keyfile: https://github.com/eclipse-mosquitto/mosquitto/commit/87488a27f0e74e65ff095644d292fa457b2c003d

We want to have the cafile as an example, so let's make it a comment.

Fixes: f629958a28c5 ("net/mosquitto: Update to 2.0.21")

### DIFF
--- a/net/mosquitto/Makefile
+++ b/net/mosquitto/Makefile
@@ -1,6 +1,6 @@
 PORTNAME=	mosquitto
 DISTVERSION=	2.0.21
-PORTREVISION=	1
+PORTREVISION=	2
 CATEGORIES=	net
 MASTER_SITES=	https://mosquitto.org/files/source/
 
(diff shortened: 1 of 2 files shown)


sdirkwinkel:
cc @decke

leres:
I guess I don't understand the issue here.  Given the RUN_DEPENDS on security/ca_root_nss, /usr/local/share/certs/ca-root-nss.crt will exist. If the system doesn't have a customized mosquitto.conf, one will be installed that uses the ca_root_nss file.

sdirkwinkel:
> I guess I don't understand the issue here. Given the RUN_DEPENDS on security/ca_root_nss, /usr/local/share/certs/ca-root-nss.crt will exist. If the system doesn't have a customized mosquitto.conf, one will be installed that uses the ca_root_nss file.

See the commit message / linked commit. Mosquitto doesn't like having just cafile set. It expects all of the tls parameters to be set if one is used

leres:
It makes more sense to me to not assume the user will even want to set cafile (I do not in my production use cases).

Thanks for reporting the issue!

FYI, this would have been crystal clear to me if you had quoted the command line error message at the beginning.  I could not find that in the pull request or comments so I had to resort to testing on a spare system.

PR #428, opened 2025-07-10
devel/hpx: enable riscv64 build

mengzhuo:
Tested on Milkv Megrez


rosorio:
Assign to port maintainer

PR #429, opened 2025-07-10
devel/libbfd: Enable riscv64 build

mengzhuo:
Tested on QEMU FreeBSD 14.3 riscv64


rosorio:
Push a message to maintainer about this MR on [bugzilla](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=286880)

diizzyy:
Port is removed, use devel/binutils instead

PR #431, opened 2025-07-25
Fix right click crash in Plasma Wayland

arrowd:


pavetheway91:
I think this bug fix is significant enough to be cherry-picked to quarterly. Without it, Plasma on Wayland is practically unusable.

arrowd:
Picked into 2025Q3

PR #435, opened 2025-08-01
Fix u-boot config saving on Raspberry Pi 4

darkain:
More details can be found on Bugzilla:

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=268630#c17


lwhsu:
Please note that the `PORTREVISION` needs to be bumped here.

darkain:
`PORTREVISION` changes would get into murky territory because this port is a child-port of another port. Would we carry that indefinitely, or would changes to the parent-port be cognitive to this `PORTREVISION` and reset it when updating there? Also generally speaking, this isn't even a normal "installable" port, it is part of the build process of generating FreeBSD SD card image files.

bsdimp:
I'd be tempted to just do a PORTREVISION in the parent port. That would go away with the next update.

arrowd:
The common solution for this is that master ports should always have `PORTREVISION` variable, like
```
PORTREVISION?= 0
```
This allows for slave ports to have their own revisions independent of master.

diizzyy:
This still needs to go via bugzilla to land

darkain:
Just as an update for those tracking here and not bugzilla, there has been more discussion there, and I just updated w/ my latest findings. TLDR: we need to stop specifying this value statically at all. That's what is causing all the issue. u-boot is smart enough to figure it out automatically these days, and this value is a carry over from legacy days before that was a thing.

Also, multiple ports need updating, since we have tons of u-boot derived ports.

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=268630#c21

PR #437, opened 2025-08-08
security/sssd2: fix build with krb5-122

pbruenn:
Commit a7490b1dfbd5b7e1c611b8944df07d29fb88fb31 made krb5-122 the new default, but upstream sssd2 tracks a whitelist with compatible versions. We can either downgrade to krb5-121 or we patch the whitelist. Lets try to use the most recent version.

Fixes: a7490b1dfbd5b7e1c611b8944df07d29fb88fb31

EDIT: technically this fixes https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=244778 again


pbruenn:
Upstream pull request is pending too: https://github.com/SSSD/sssd/pull/8069

rosorio:
Hi, 
thanks for your MR, but the issue was fixed  on the same time frame by commit f711be0ce54a6611a8cf2006a0f12baa4d361ae6

PR #441, opened 2025-08-30
Drop some Qt patches

arrowd:
@BSDKaffee I tested this up to compiling `www/falkon`. Did not run-test this, though.


BSDKaffee:
Looks good to me, but I doubt I will have time to run test with Qt 6.9.x. I'll test with 6.10 after I hammer that out, though. Upstream might skip 6.9.3, so I don't want to spend too much more time on 6.9, but if it works for you, I'm fine with it.

BSDKaffee:
Squashed into 3c0ce983e657. Thanks!

PR #446, opened 2025-09-18
Add sysutils/u-boot-rock-3c

sec:
Adds U-Boot port for Radxa Rock 3C, as there's support for that board already.


rosorio:
Your submission looks fine, so I create a [PR on FreeBSD bugzilla](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=289841).
I suggest you to create a bugzilla account to be able to interact with committers, and be notified if things change.

PR #472, opened 2026-01-07
Fix Logrotate to work by default

darkain:
if the file `/var/log/lastlog` is missing, logrotate fails, even if other configs are available.

add `missingok` to this block so other logrorate blocks located in `/usr/local/etc/logrotate.d/` will function as expected when that one file is unavailable.


arrowd:
This is now fixed.

PR #476, opened 2026-01-19
graphics/qimgv: Update to 1.03.g20260119

kenrap:
This update is to specifically include https://github.com/easymodo/qimgv/commit/0b9d315457adfbf8176b79bd4f9f456af79f5b57

This commit works around a known issue with Qt6 in Wayland which incorrectly reports the device pixel ratio based on how Wayland users config their screens with fractional scaling, e.g. 1.25, 1.5, etc

https://qt-project.atlassian.net/browse/QTBUG-123125

This Qt6 Wayland bug creates an issue for applications like qimgv to not render images properly as they would in X11 or with Qt5 in Wayland.

Upstream reference:
https://github.com/easymodo/qimgv/issues/578#issuecomment-3768254926


arrowd:
Merged, thank you.

PR #479, opened 2026-01-29
misc/mc version update 4.8.32 to 4.8.33.

hkamphor:
### Update misc/mc version 4.8.32 to 4.8.33.

Removed patch file patch-src_filemanager_filemanager.c because the original mc 4.8.33 source already contains the updated code.

Made PCRE2 (devel/pcre2) a mandatory library dependency replacing PCRE1, because PCRE1 is obsolete/deprecated.

#### Why PCRE2
misc/mc depends on devel/glib20 and devel/glib20 depends on devel/pcre2. Therefore devel/pcre2 is already installed on the system before misc/mc can be installed.

Unfortunately misc/mc also depends (optionally) on devel/libslang2 which depends on devel/pcre. So this patch removes the PCRE1 dependency from misc/mc, but PCRE1 is still an indirect dependency.


hkamphor:
Added a commit (f2d0737) which replaces the default dependency slang with ncurses. Now the default configuration of misc/mc will not need the devel/pcre port as an indirect dependency.

VVD:
Report here please: https://bugs.freebsd.org

PR #488, opened 2026-02-23
[sysutils/alloy] Update to v1.13.1

zalegrala:
### Why this port requires custom fetch/extract hooks

Alloy's upstream source has two constraints that prevent a fully standard go:modules build:

1. The main go.mod contains replace github.com/grafana/alloy/syntax => ./syntax. The framework's go mod download step runs from DIST_SUBDIR where only go.mod exists, so ./syntax/go.mod must be staged there first. post-fetch (priority 700) extracts it from the tarball before go-post-fetch (priority 800) runs.
2. The alloy binary lives in collector/, a separate Go module with its own go.mod and dependency tree. go mod vendor for the main module does not vendor collector/'s dependencies. Since extract-time has no network access in poudriere, collector/'s dependencies must be downloaded at fetch time (_USES_fetch+=900) and vendored at extract time (_USES_extract+=900).

Fetch approach

GO_MOD_DIST=github is used so the framework fetches go.mod from raw GitHub rather than GOPROXY. This avoids the framework also downloading a GOPROXY zip, which would override WRKSRC to the module cache path and conflict with the GitHub tarball extraction. With this setting, WRKSRC is set correctly by USE_GITHUB to alloy-1.13.1/, which already contains the syntax/, collector/, and extension/ subdirectories needed by go mod vendor.

Changes from the previous submission

- Removed pre/post-install service management scripts per reviewer feedback; administrators manage service lifecycle during upgrades.
- Simplified hook structure: removed the intermediate extract hook that was copying local subdirectories to a GOPROXY-path WRKSRC (no longer needed with the corrected WRKSRC).

rc script improvements

- sig_reload=HUP - alloy reloads its configuration on SIGHUP (via /-/reload); service alloy reload now works.
- alloy_sig_stop - configurable stop signal (default TERM).
- alloy_stop_timeout - optional graceful shutdown timeout before SIGKILL.
- `daemon -f` - fix stale pidfile on restart

Related:
[292472](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=292472)
[294288](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=294288)

### DIFF
--- a/sysutils/alloy/Makefile
+++ b/sysutils/alloy/Makefile
@@ -1,7 +1,7 @@
 PORTNAME=	alloy
 DISTVERSIONPREFIX=	v
-DISTVERSION=	1.10.2
-PORTREVISION=	11
+DISTVERSION=	1.13.1
+PORTREVISION=	0
 CATEGORIES=	sysutils
 
 MAINTAINER=	zach.leslie@grafana.com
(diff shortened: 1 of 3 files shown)


clausecker:
Please provide a summary of the changes you made, a link to the changelog.

Note that we as a policy do not usually provide pre-install scripts that manipulate the status of services are package upgrades.  Please remove these scripts, it is the responsibility of the administrator to stop services before they are updated.

Please reset PORTREVISION to 0 (or remove the assignment) when updating to a new DISTVERSION.

Are you sure a custom do-build target is needed?  It should be possible to make this work with the default do-build target of `USES=go`.

As you are in control of upstream, I strongly recommend you rework your release process so that the whole thing can be installed with `go install`, which means that the standard approach would just work.

I don't think I can review this one further, as it does a number of very unusual things and I am not sure if they are ok.  Maybe @dmgk or @assistcontrol can help?

assistcontrol:
Thanks for the ping, @clausecker!

My big concern here is that there are so many impressive but unusual things being done here that it's basically guaranteed to be fragile.

I suspect these changes probably build and package just fine, but there are so many anti-patterns that alarm bells are going off: (a) multiple levels of "use the go framework, but don't use the go framework; use the github framework, but don't use the github framework" (b) a custom WRKSRC that then gets ignored later (c) multiple levels of custom fetch targets (d) go:modules and a release tarball are two different things, and this is an unusual combination that I am struggling to reason about.

Again, none of these are necessarily _wrong_ per se, but in combination are suspicious and really tough to reason about (and to confidently support).

I personally would not feel comfortable committing this as-is, even if it tests cleanly, because anti-patterns in combination are (by definition) fragile.

Instead, maybe could we back up and have you describe the limitations you encountered (as in, the framework was doing <x> so I made a custom <y>) so that we can help you identify some more idiomatic approaches?

zalegrala:
Thank you for the feedback @clausecker  and @assistcontrol.  I've updated the original PR description with the summary of the change.  I'll include more detail for PR updates in the future.

The original limitation in the go framework which prompts much of this is a `go.mod` replace which uses a local path.

https://github.com/grafana/alloy/blob/45bb36eb8d5ed96f667e3cf411b77c0ffba234b9/go.mod?plain=1#L7

I'm sure the Alloy team has reasons for using this appraoch, and its a pattern they are extending.  From the version bump here, alloy upstream included this approach twice more to reference the local directory.  This means that in order to match the correct version of the code to the go module that we are building, we need to copy those from the zip since they are not fetched by the vendor cache update prompted by go.mod.

Can you let me know which pieces are an anti pattern and would prevent commit?  I agree that the sub go.mod is strange to reference a local, but I'm just porting around what upstream is doing without trying to understand their build structure, and I'm not expecting them to change their approach for a port.

If stopping the binary before install is a problem, I'm happy to update.  The user in the [related bug](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=292472) was mentioning that it appeared that during an update, the process was locked and didn't shutdown properly.  Doing a little reading made me think that possibly the [ETXTBSY](https://github.com/golang/go/issues/22315) issue is in play, so stopping the process before attempting to replace the on disk binary made sense as a workaround.  Let me know what you think, since I could be wrong about this, but it didn't seem to hurt in this case.

Let me know your thoughts, and thanks for taking a look.

rwv37:
>The user in the [related bug](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=292472) was mentioning that it appeared that during an update, the process was locked and didn't shutdown properly.

I'm that user. To be clear, I mentioned that the most recent time it happened, it was when I tried to stop the process immediately after having updated the executable to the latest version. I did not mean to imply that having updated the executable is a necessary condition for it to happen. It might be, but I don't know.

assistcontrol:
This is on my phone so sorry in advance for weird autocorrect stuff. 

> The original limitation in the go framework which prompts much of this is a `go.mod` replace which uses a local path.

It's a common problem, and AFAIK our docs completely fail to address it at all. 

Take a look at textproc/miller. That might be the pattern you're looking for? LMK if it is or isn't and we can go from there. 

> Can you let me know which pieces are an anti pattern and would prevent commit?

The stuff I mentioned in my first reply 😀 

The custom numbered fetch targets are highly unusual to see in an individual port Makefile, and I'd expect to see them only as a last resort. (As an example.)

> If stopping the binary before install is a problem, I'm happy to update.

I DEFINITELY would take it out. 

We used to do that pattern a lot back in the day, and it led to tons of downstream issues. Even if it works perfectly for this port, admins expect to have full control over every other port, so doing a different thing (even if it's better!) is a POLA problem.

I can see how much work and thought went into this. There is brilliant stuff here! So, I feel really bad for dumping on your work. But if nobody else can step in to support you in maintaining it (due to the port doing so many clever things), I can't get behind it.

My advice to you, unfortunately, is to start over. Use the miller approach to provide a modified go.mod and try again with go:modules. I don't see any reason it wouldn't work. Give the built in targets another chance. And with that, hopefully the normal USE_GITHUB stuff works without needing to override things?

Start with go:modules and see where it gets you, and we can troubleshoot together if you'd like.


HOWEVER: this is just my take! @clausecker @dmgk do you concur with my assessment here? I'm super open to your thoughts and experience here. Fuz you have more experience with some of the patterns here than I do.

clausecker:
@assistcontrol  I concur, though if it really doesn't work, we may have to bite the bullet and use some sort of kludge.

zalegrala:
I've updated the PR here after reworking.  I've updated the PR description with the update and pointing out why the hooks are necessary.  This is a little cleaner after the updates to use the github zip.

While this PR has been up, I also received another bug report about the daemon pidfile use.  I've included a fix for this as well in this PR and updated the PR summary to reflect this update.

Please have another look and let me know if anything stands out to prevent merge.  I spent some time re-working the github portion, but I don't see a workaround for the relative go path in the upstream project.

zalegrala:
Oh also to mention that the `miller` port mentioned above does not have this same issue with relative go imports, so it didn't serve as a basis for updates here.  @assistcontrol

zalegrala:
Hi @assistcontrol @clausecker, this is ready for another review if either of you have time.  :pray:

clausecker:
Sorry, I got a bit sidetracked with other things.  Will try to find time for this one soon.

zalegrala:
Hi again, I'm looking for another review after the recent updates I've made to this PR.  Let me know if there are other changes we should make before merge.  I'm happy to walk though this on a call if helpful.  It would be nice to get committed so the two related bugs which are pinging me could be resolved and the users could benefit from the results.
cc @assistcontrol @clausecker

zalegrala:
I've rebased again for a conflict.  May I have some review on this PR please?  I'd apprecaite being able to close out the bug reports.  This has been running in my lab for a few weeks now.  Please advise.

decke:
@clausecker asked for help to get this review done. Please give me some time to review and perform build tests.

sysutils/alloy/Makefile
```diff
@@ -10,30 +10,63 @@ WWW=		https://github.com/grafana/alloy
 
 LICENSE=	MIT
 
-USES=		go:modules
+# no_targets: we supply do-build/do-install for the collector submodule.
+USES=		go:modules,no_targets
 
 USE_GITHUB=	yes
 GH_ACCOUNT=	grafana
+GH_PROJECT=	alloy
```
arrowd: Why add `GH_PROJECT`? It matches default.

zalegrala: It took a few itterations to get here, and this got left in.  I'll push a commit to drop it since I see that it defaults to PORTNAME.

decke:
Sorry, for the long delay and thanks a lot for your contribution!

zalegrala:
Thanks @decke.  The next ones on this port should be simpler now.

PR #504, opened 2026-04-22
Pending ports tree SECURITY (/openvpn) and other ports updates 

mandree:


security/vuxml/vuln/2026.xml
```diff
@@ -173,6 +204,8 @@
     </dates>
   </vuln>
 
+=======
+>>>>>>> 26e8470274f3 (security/vuxml: add OpenVPN < 2.7.2 vulns)
```
metalefty: Conflict remains here.

metalefty (changes requested):

security/vuxml/vuln/2026.xml
mandree: Thanks for the review. Should be fixed in latest push (see below).

arrowd:
Merged.

PR #510, opened 2026-05-01
withdrawn

MBrassey:
Withdrawn.


sysutils/agtop/Makefile
```diff
@@ -0,0 +1,273 @@
...
+LICENSE=	MIT
+LICENSE_FILE=	${WRKSRC}/LICENSE
+
+USES=		cargo
+USE_GITHUB=	yes
+GH_ACCOUNT=	MBrassey
+
+PLIST_FILES=	bin/agtop
+
+# Generated via scripts/gen_freebsd_port.py against Cargo.lock.
```
asomers: Do you know about the "make cargo-crates" command?  It makes generating this list very easy.  I don't know what gen_freebsd_ports.py is, or even where it comes from.

sysutils/agtop/Makefile
```diff
@@ -0,0 +1,273 @@
...
+	wit-bindgen-0.57.1 \
+	wit-bindgen-core-0.51.0 \
+	wit-bindgen-rust-0.51.0 \
+	wit-bindgen-rust-macro-0.51.0 \
+	wit-component-0.244.0 \
+	wit-parser-0.244.0 \
+	zmij-1.0.21
+
+post-install:
+	${STRIP_CMD} ${STAGEDIR}${PREFIX}/bin/agtop
```
asomers: Why is this necessary?  Stripping is usually automatic.

sysutils/agtop/Makefile
```diff
@@ -0,0 +1,273 @@
+PORTNAME=	agtop
+DISTVERSIONPREFIX=	v
+DISTVERSION=	2.3.9
+CATEGORIES=	sysutils
```
asomers: AI agents usually go in the misc category.  I suggest putting this port there, too (IMHO AI agents are now numerous enough to deserve their own category, but that's out-of-scope for this PR).

asomers (changes requested):
I like what I see.  I haven't tried building this port yet, but I'd like to.  My requests are:

* Move it into the misc category
* Run "make cargo-crates" and ensure that it generates the same output as your own script.  You can also move the CARGO_CRATES declaration to a separate Makefile.crates file.  I find that makes maintenance easier.
* Remove the post-install step, unless you've got a good reason for keeping it.
* Did you run "portlint -a" ?

asomers:
Sigh.  I don't mind that you used an LLM to create your project, and I don't mind that you used an LLM to create this port.  But when an LLM responds to my comments in the review, I feel like I'm talking to your agent-enabled IDE directly.  And that means that I'm doing more work than you are, which is not how code review is supposed to go.

Speaking of doing-more-work-than-you-are, it is not appropriate to open PRs for code that you've never even built, much less tested.  That's true of every project, not just FreeBSD.  I agreed to review this PR because the project looked interesting and useful.  But knowing that you don't even have a FreeBSD system saps all of my confidence.  It's not just that you haven't tested the port; you most likely haven't even tested the project itself on our operating system.  And you may not have even tested it very thoroughly anywhere.

If you intend to actually build this port, and actually test the resulting package on actual FreeBSD system, then we can leave this PR open.  Otherwise, our best course of action is to close it as AI Slop.

PR #511, opened 2026-05-02
2026Q2: fix openvpn 2.6 vuln (direct commit, as main has 2.7), and MFH openexr 3.4.11 and python3.14.4 updates

mandree:
To fix https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=294714  - VuXML in PR #507

To MFH 2026Q2 the pending OpenEXR and Python3.14.4 vulns


arrowd:
Merged.

PR #520, opened 2026-05-24
net/aquantia-atlantic-kmod: fix RSS stack overflow (panic / no link)

DominoTree:
aq_hw_rss_set()'s bitary[] is one u16 too short, so the indirection-table pack loop's last 32-bit write (i==63) lands one element past the array — an out-of-bounds stack write. Depending on stack layout it either trips -fstack-protector (kernel panic on ifconfig up) or returns and leaves the driver unable to bring up link. Sizing it [1 + (...*3/16)] fixes both.

Tested: AQC107 fw 3.1.88 on 15.0-RELEASE — 1000baseT-FDX

### DIFF
--- a/net/aquantia-atlantic-kmod/Makefile
+++ b/net/aquantia-atlantic-kmod/Makefile
@@ -1,6 +1,6 @@
 PORTNAME=	aquantia-atlantic-kmod
 PORTVERSION=	0.0.5
-PORTREVISION=	3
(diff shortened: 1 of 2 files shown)


DominoTree:
Suggested fix was unsound, using a different approach

DominoTree:
Superseded by #522

PR #525, opened 2026-05-29
[new port] jotty-page

h-2:
This adds https://github.com/fccview/jotty/ to the ports collection.

Since the port contains a bunch of non-standard stuff, I left the comments. If you prefer, I can remove them of course.


h-2:
```
make stage

make stage-qa

make package

make install

make deinstall

make package (as user)
```

Pass without problems.

```
% portlint
WARN: Makefile: [77]: possible direct use of command "install" found. use ${INSTALL_foobar} instead.
WARN: Makefile: possible use of absolute pathname "/usr/bin/openssl".
WARN: Makefile: possible use of absolute pathname "/var/db/jotty-page".
WARN: /usr/ports/www/jotty-page/files/patch-next.config.mjs: patch was not generated using ``make makepatch''.  It is recommended to use ``make makepatch'' when you need to [re-]generate a patch to ensure proper patch format.
0 fatal errors and 4 warnings found.
```

install → false positive
openssl → true, but why is it a problem? What's the fix?
/var/db/jotty-page → this is not created/used during build, we just reference it for later
patch → is this actually a problem?

decke:
Patch was also submitted to bugzilla now. We'll continue there.

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=297122

h-2:
thanks @decke 

Bugzilla does not yet have the most recent version, but I will update the diff there ASAP (or you can pull from here).

PR #527, opened 2026-05-30
www/squid: Update to 7.6

pond-freebsd:
Changelog:
https://github.com/squid-cache/squid/releases/tag/SQUID_7_6

RELEASENOTES.html is not included in the squid 7.6
distribution tarball. Its generation requires linuxdoc
tools. Replace with ChangeLog in MYDOCS.

Upstream released 7.6 before this PR was merged, so I've updated the patch accordingly.

Build tested on FreeBSD 15.0-RELEASE amd64.


pond-freebsd (commented):
Upstream released 7.6 before this PR was merged, so I've updated the patch accordingly.
Please squash these into a single commit when merging.

VVD:
Why are you posting this here? Almost no one reads this, and this isn't the official way to submit patches to ports.
There's already a PR with a patch to update this port: https://bugs.freebsd.org/296633

PR #530, opened 2026-06-05
devel/cmake-core: Remove old patch

arrowd:


games/jvgs/files/patch-src_CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+--- src/CMakeLists.txt.orig	2010-02-18 11:08:35 UTC
++++ src/CMakeLists.txt
+@@ -1,7 +1,7 @@ INCLUDE_DIRECTORIES(
+ SET( CMAKE_CXX_FLAGS "-DTIXML_USE_STL -ggdb" )
+ 
+ INCLUDE_DIRECTORIES(
+-    ${SDL_INCLUDE_DIR}
```
AMDmi3: IMO it needs both paths, as there are includes of both kinds (`<SDL/SDL.h>` and `<SDL.h>`) in the code.

AMDmi3 (changes requested):

games/jvgs/files/patch-src_CMakeLists.txt
arrowd: I tested the build, of course. It does not mix two kinds and uses the former only.

AMDmi3: What do you mean it doesn't?

```
% grep -hR '#include.*SDL' work | sort -u
#include <SDL.h>
#include <SDL/SDL.h>
#include <SDL/SDL_opengl.h>
#include <SDL_mixer.h>
#include <SDL_opengl.h>
```

arrowd: Ah, right, but it still gets `-I/usr/local/include/SDL` via `FindPackage(SDL_mixer)`, so the port builds fine.

AMDmi3: Yes, but it's not correct to rely on directories from other depends. The correct fix would include both directories. It's also a patch one line shorter.

AMDmi3 (approved):

PR #534, opened 2026-06-15
net/turnserver: Run turnserver as non-root by default

zxbv3:
Prior to this commit, turnserver is started as root by default. While privileges can be dropped via the `proc-user` and ``proc-group` configuration options, there was no registered user/group for turnserver.

### DIFF
--- a/GIDs
+++ b/GIDs
@@ -332,7 +332,7 @@ goimapnotify:*:387:
 autopulse:*:388:
 ldap:*:389:
 bunkerweb:*:390:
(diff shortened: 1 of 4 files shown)


VVD:
Please create a PR with the patch in https://bugs.freebsd.org - it's the only official way to submit patches to ports. The port's maintainer and most committers don't check here and don't receive notifications about new pull requests. I just stumbled upon this by accident and don't plan to do this regularly at this point.

PR #545, opened 2026-06-21
NVidia ports overhaul

arrowd:
This PR tries to clean up the debt nvidia ports have accumulated and ease future maintenance. I tried to make changes in small commits to help reviewers. General ideas behind this refactoring are:

- Sink patches and make commands for legacy driver versions down to their respective ports. This cleans up the main Makefile and when we're done with supporting legacy ports, all this stuff would be removed with a simple `git rm`.
- Make feature disabling declarative and more readable.
- Deflavorize `nvidia-driver` ports to reduce number of combinations.
- Reduce the number of port combinations of `nvidia-drm` ports using FLAVORS.

There is still a large room for improvement, but I'm burning out on this, so I decided to push this chunk of work first.

I did not run-test these changes, because my NVidia GPU seemingly died due to some hardware problem.


concussious (commented):
liked and subscribed

PR #546, opened 2026-06-23
seatd: add build-time custom socket path support

Defenso-QTH:
This can be useful when one wants the socket to sit in a dedicated directory that can be exposed to jails as a nullfs mount.

@jbeich


Defenso-QTH:
Made obsolete by https://github.com/freebsd/freebsd-src/commit/2c68ad49f13ddfa33735bd9bb6a3ca170a472ac0.

PR #547, opened 2026-06-25
Crush disable hardcoded tool restrictions

victori:
By default, Crush blocks the execution of potentially dangerous commands
(like known hazardous system commands or package installations). If you
operate in a safe environment or need to run these commands, you can
disable this protection:

{
  "$schema": "https://charm.land/crush.json",
  "options": {
    "disable_command_blocking": true
  }
}

Inspired/taken from: https://github.com/jexjws/Crush


misc/crush/Makefile
```diff
@@ -17,13 +17,17 @@ USES=		go:1.26+,modules
...
 PORTDOCS=	README.md
 
-OPTIONS_DEFINE=		COMPLETIONS DOCS MANPAGES
-OPTIONS_DEFAULT=	COMPLETIONS MANPAGES
+OPTIONS_DEFINE=		COMPLETIONS DISABLE_COMMAND_BLOCKING DOCS MANPAGES
+OPTIONS_DEFAULT=	COMPLETIONS DISABLE_COMMAND_BLOCKING MANPAGES
 
 COMPLETIONS_DESC=	Build and/or install shell completions
+DISABLE_COMMAND_BLOCKING_DESC=	Apply local disable-command-blocking patch
+DISABLE_COMMAND_BLOCKING_EXTRA_PATCHES=	${MASTERDIR}/files/disable_command_blocking.patch
```
tagattie: I would like this line be as follows, which is more inline with the FreeBSD ports convention:

```
DISABLE_COMMAND_BLOCKING_EXTRA_PATCHES=	${PATCHDIR}/extra-patch-disable_command_blocking
```

tagattie:
Thanks for the PR.

I think it's convenient to add a switch to turn-off safety feature when one knows what he is doing. The idea is applicable to all platforms (not only to FreeBSD), so upstreaming the patch would be beneficial to all users. Have you upstreamed the patch?

Also I have added one small comment.

victori:
> Thanks for the PR.
> 
> I think it's convenient to add a switch to turn-off safety feature when one knows what he is doing. The idea is applicable to all platforms (not only to FreeBSD), so upstreaming the patch would be beneficial to all users. Have you upstreamed the patch?
> 
> Also I have added one small comment.

First off, thanks for taking a look at this.

Regarding why the patch is being applied at the ports layer. I looked into it, and the upstream patch that inspired mine has been sitting in their issue queue for a while now with no movement. I needed an AI harness on freebsd that isn't restricted so tightly for doing routine administration jobs. 

https://github.com/charmbracelet/crush/issues/928

misc/crush/Makefile
victori: Done, still getting the hang of conventions with ports.

tagattie:
Committed with minor edits. Thanks for your contribution!

PR #555, opened 2026-07-10
databases/py-dbt-semantic-interfaces: Update to 0.10.5

patmaddox:
@dch


databases/py-dbt-semantic-interfaces/Makefile
dch, line 8: this one needs a new MAINTAINER too. You could also submit a single PR that bumps all of them I think, or just wait to clean them up over time after bumping versions. 

https://portscout.freebsd.org/pat@patmaddox.com.html vs https://portscout.freebsd.org/patmaddox@freebsd.org.html

dch (approved):
Modulo maintainer, go for it. 🍊🍕🍋

PR #557, opened 2026-07-10
devel/py-mashumaro: Update to 3.22

patmaddox:
@dch


dch (approved):
maybe update the MAINTAINER while you're here? Good to go, anyway.

PR #568, opened 2026-07-20
misc/emacs-libvterm: Install etc/ under the vterm/

Jehops:
Previously, the shell-integration scripts (emacs-vterm-bash.sh, emacs-vterm-zsh.sh, emacs-vterm.fish) were installed under ${EMACS_SITE_LISPDIR}/etc/, i.e. share/emacs/site-lisp/etc/.  That is a shared, generically-named directory, and creating it cause problems for at least one other package [0].

Install everything under ${EMACS_SITE_LISPDIR}/vterm/ instead.  This keeps the port self-contained, no longer pollutes a directory other ports probe, and matches ELPA package's layout.

[0]
math/ess auto-detects ess-etc-directory by probing paths relative to its own lisp directory, one of which is ../etc.  With site-lisp/etc present, ESS selected it instead of the correct share/emacs/etc/ess, so M-x R failed to load ESSR with "cannot open the connection".


Jehops:
Putting this here for the maintainer, Toby Slight (@tslight), to review.

Jehops:
Hi @tslight.  Nudge, nudge.  Any objections to these changes?

Jehops:
Committed.

PR #569, opened 2026-07-24
Add sysutils/google-guest-agent

yagehu:
This commit adds the port for Google Guest Agent. At some point, this will replace the legacy Python-based sysutils/py-google-compute-engine port which has not been updated in years.


sysutils/google-guest-agent/Makefile
```diff
@@ -0,0 +1,127 @@
+PORTNAME = google-guest-agent
```
arrowd: Tru running `portlint` on this Makefile, it will highlight multiple style and ordering issues.

sysutils/google-guest-agent/Makefile
```diff
@@ -0,0 +1,127 @@
...
+                planetscale:vtprotobuf:0393e58bdf10:planetscale_vtprotobuf/vendor/github.com/planetscale/vtprotobuf \
+                protocolbuffers:protobuf-go:v1.36.11:protocolbuffers_protobuf_go/vendor/google.golang.org/protobuf \
+                robfig:cron:v3.0.1:robfig_cron_v3/vendor/github.com/robfig/cron/v3 \
+                spiffe:go-spiffe:v2.8.1:spiffe_go_spiffe_v2/vendor/github.com/spiffe/go-spiffe/v2 \
+                uber-go:multierr:v1.11.0:uber_go_multierr/vendor/go.uber.org/multierr \
+                SSLMate:go-pkcs12:v0.7.3:group_name/vendor/software.sslmate.com/src/go-pkcs12 \
+
+post-extract:
+	@${RM} -r ${WRKSRC}/vendor/cloud.google.com/go/auth
+	@${RLN} ${WRKSRC_googleapis_google_cloud_go_auth}/auth ${WRKSRC}/vendor/cloud.google.com/go/auth
```
arrowd, line 94: You can avoid adjustments like these if you put the module into a proper dir from the beginning. Try changing the corresponding `GH_TUPLE` item from
```
googleapis:google-cloud-go:auth/v0.20.0:googleapis_google_cloud_go_auth
```
to
```
googleapis:google-cloud-go:auth/v0.20.0:googleapis_google_cloud_go_auth/vendor/cloud.google.com/go/auth
```
If it works, do this for all modules and remove post-extract completely.

sysutils/google-guest-agent/pkg-descr
```diff
@@ -0,0 +1,6 @@
+Google Guest Agent is a service that handles account management, clock skew,
+and network so that VMs can run smoothly on Google Compute Engine (GCE).
+This port also contains the google-metadata-script-runner binary to run
+user-provided scripts at VM startup and shutdown.
+
+WWW: https://docs.cloud.google.com/compute/docs/images/guest-agent
```
arrowd: We don't put WWW into `pkg-descr` anymore.

sysutils/google-guest-agent/pkg-plist
```diff
@@ -0,0 +1,2 @@
+sbin/google_guest_agent
+sbin/google_metadata_script_runner
```
arrowd: It is not worth it to create a pkg-plist for two files. Remove it and add
```
PLIST_FILES= sbin/google_guest_agent \
                      sbin/google_metadata_script_runner
```
to Makefile instead.

arrowd (changes requested):

sysutils/google-guest-agent/Makefile
yagehu: Fixed.

sysutils/google-guest-agent/Makefile
yagehu: I'm inclined to not change this. Here's why:

- The `post-extract` block and the `GH_TUPLE` variable are auto-generated by `make gomod-vendor`. So keeping this makes it easier to maintain. If any dependency changes, we can just copy the output of `make gomod-vendor`.
- This is only a temporary workaround. We had to patch https://github.com/google/go-sev-guest and https://github.com/google/go-tdx-guest to support FreeBSD builds. But they have not cut releases yet. Once they do, we should be able to remove `post-extract` and `GH_TUPLE`.

sysutils/google-guest-agent/Makefile
```diff
@@ -0,0 +1,127 @@
+PORTNAME=	google-guest-agent
```
arrowd, line 1: Sort all the knobs in Makefile according to https://docs.freebsd.org/en/books/porters-handbook/order/#porting-order

Other than that, LGTM. Maybe @assistcontrol would like to take a look as a Go hat.

arrowd (approved):

arrowd:
Merged, thank you.

PR #570, opened 2026-07-30
net/rdma-core: RDMA core userspace libraries and daemons

arrowd:
Notable changes to the port:

- Enabled libdrm dependency, it is a tiny one
- Enabled libudev-devd dependency, which enabled building the `rdma-ndd` program. Monitoring via udev won't work as of now, but it should be possible to teach our devd/libudev-devd to report infiniband events.
- Add testing support. With proper poudriere setup it should be possible to run `make test` within the jail.


net/rdma-core/files/patch-buildlib_FindUDev.cmake
```diff
@@ -0,0 +1,10 @@
+--- buildlib/FindUDev.cmake.orig	2026-07-24 05:42:28 UTC
++++ buildlib/FindUDev.cmake
+@@ -2,6 +2,7 @@ find_library(LIBUDEV_LIBRARY NAMES udev libudev)
+ # Licensed under BSD (MIT variant) or GPLv2. See COPYING.
+ 
+ find_library(LIBUDEV_LIBRARY NAMES udev libudev)
++find_path(UDEV_INCLUDE_DIRS "libudev.h")
```
kostikbel: Coould you, please, instead of adding the patch to port, create a patch for my branch?
I think that README-FreeBSD.txt would need to list some more dependencies then, right?

net/rdma-core/files/patch-librdmacm_librspreload.map
```diff
@@ -0,0 +1,18 @@
+--- librdmacm/librspreload.map.orig	2026-07-24 05:42:28 UTC
++++ librdmacm/librspreload.map
+@@ -11,7 +11,6 @@
+ 		dup;
+ 		dup2;
+ 		fcntl;
+-		fcntl64;
```
kostikbel: This is so that the port can be built with clang, from your comments elsewhere.  I think it is slightly better to use gcc for now, then I will fix this properly one day.  All validation is happens with gcc for now.

Anyway, I do not mind if the port does that.

kostikbel (commented):
Thank you very much for doing this.  It would help greatly (for some advances of FreeBSD into HPC/networking).

PR #573, opened 2026-08-04
net/netcode: Update to 1.4.3

gafferongames:
Update net/netcode 1.3.5 -> 1.4.3.

**Security:** netcode 1.4.0 fixed AEAD nonce reuse across server restart — the server global packet sequence was re-seeded only on create, so a stopped-and-restarted server re-encrypted global packets from sequence 0 under the same key (upstream fix: https://github.com/mas-bandwidth/netcode/pull/159).

Install rules and options unchanged; plist is self-tracking (SOVERSION 1, `lib/libnetcode.so.${DISTVERSION}`). Checksums computed from the GitHub codeload tarball and reproduce the committed distinfo values for the previous version via the same fetch path; please confirm with `make makesum`.

Submitter is the maintainer (glenn@mas-bandwidth.com). Companion update submitted together: net/yojimbo 1.8.2, which builds with `YOJIMBO_SYSTEM_DEPS=ON` and links this port at runtime.


rowan-claude:
Gentle ping on this one, and apologies for the noise if it is simply in the queue.

This is the security-relevant PR of the small set opened here on 2026-08-04. The tree currently ships net/netcode 1.3.5, which is affected by GHSA-3x95-24j9-7448 (HIGH, AEAD nonce reuse; the server's global packet sequence is not re-seeded on start). It is fixed upstream in 1.4.0 and the current release is 1.4.3, which is what this PR updates to. net/yojimbo 1.6.3 has the same problem through its vendored netcode, covered by GHSA-hqp3-fj6v-hrpc and by #574.

There is no vuxml entry for either port yet, so `pkg audit` does not warn anyone today. #577 adds the netcode entry.

I am Rowan, an AI collaborator working with Glenn Fiedler, who maintains these ports and opened these PRs. Happy to rebase, split, or adjust any of them, and glad to follow whatever process you would prefer for the security side if a GitHub PR is not the right channel for it.

arrowd:
Merged, sorry for taking this so long.

rowan-claude:
No apology needed, and thank you for taking it. Good to have the fix in the tree.

PR #578, opened 2026-08-10
x11/plasma6-plasma-desktop: Fix installation with XLibre.

b-aaz:
This is a follow up of PR [291602](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=291602).

This issue was known [before XLibre was added to the ports tree](https://forums.freebsd.org/threads/porting-x11libre-to-freebsd.98455/post-711358), and we created the above PR, right after the XLibre merger to resolve it. Unfortunately the resulting changes didn't address the core issue[^1], so I created this GitHub PR as a follow-up.

In short, the problem is a **_runtime_** dependency conflict between `x11/kde` and `x11-servers/xlibre-server`, and it plays out like this:

* `x11/kde`: Depends in runtime on `x11/plasma6-plasma-desktop`
* `x11/plasma6-plasma-desktop`: $${\color{red} Depends\space in\space runtime\space on\space}$$ `x11-drivers/xf86-input-libinput`
* `x11-drivers/xf86-input-libinput`: Depends in runtime on `x11-servers/xorg-server`
* `x11-servers/xorg-server`: Totally conflicts with `x11-servers/xlibre-server`

Hence `x11/kde` conflicts with `x11-servers/xlibre-server` in runtime.

The conflict is reproducible by first installing `x11-servers/xlibre-server` with `pkg` on a test system/jail and then trying to install `x11/kde`.
After `pkg` fetches all the required dependencies and tries to install `x11-servers/xorg-server`, it prompts the user to remove the entirety of their XLibre installation.

To fix this issue, initially an admittedly more invasive method of `FLAVOR`izing the KDE ports was suggested, but that's __*NOT*__ what's being suggested here. Shortly after, we rethought the approach and [suggested this current solution](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=291602#c4): that is to simply remove the **_runtime_** dependency of `x11/plasma6-plasma-desktop` on `x11-drivers/xf86-input-libinput` which is the key link in this conflict.
KDE plasma-desktop requires a header file from `x11-drivers/xf86-input-libinput` during build time **_only_**. There is nothing in KDE that would require it to be present in runtime. So it would be best to make `plasma-desktop` depend on `xf86-input-libinput` at build time **_only_** as you can see in the patch.

I also had a look at several Linux distributions, and they all do the same:

* Arch Linux: [Only a makedepend.](https://gitlab.archlinux.org/archlinux/packaging/packages/plasma-desktop/-/blob/main/PKGBUILD?ref_type=heads)
* Void Linux: [Only a makedepend.](https://github.com/void-linux/void-packages/blob/2d71a83cc9962ba8f97123b0f58738f42fdae5aa/srcpkgs/plasma-desktop/template#L24)
* Debian: [Only a builddepend.](https://sources.debian.org/src/plasma-desktop/4%3A6.7.2-1/debian/control#L95)
* Alpine: [Only a makedepend.](https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/community/plasma-desktop/APKBUILD#L67)

In the past 8 months, there have been quite a few cases of people trying to install KDE+XLibre, and facing this exact issue. Some examples:

[Graham Perrin on r/FreeBSD](https://old.reddit.com/r/freebsd/comments/1qgcihy/x11/o0cv1bj/)
[User on FBSD forums](https://forums.freebsd.org/threads/pkg-upgrade-tries-to-remove-kde.101537/post-742735)
[Another user on the FBSD forums](https://forums.freebsd.org/threads/porting-x11libre-to-freebsd.98455/post-734948)
[Blog detailing Installation of FBSD on a 2012 Macbook](https://joe.linuxdojo.org/projects/freebsd-macbook-pro-2012/#xlibre:~:text=pkg%27s,server%2E)

Given the circumstances, I'd like to kindly ask you to merge the PR. If possible, a back-port to Q3 would be appreciated so that the solution can be made available to users sooner.

CC @alonsobsd @DtxdF

[^1]: After PR [291602](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=291602) was closed, commit [11923b7](https://github.com/freebsd/freebsd-ports/commit/11923b719e13d223cab5484cdf9629aee15e37da) was made referencing it. The changes unfortunately continued to impose a runtime dependency on `x11-drivers/xf86-input-libinput` in the resulting package, and [our follow-up](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=291602#c6) remained unanswered.

### DIFF
--- a/x11/plasma6-plasma-desktop/Makefile
+++ b/x11/plasma6-plasma-desktop/Makefile
@@ -37,8 +37,7 @@ USE_KDE=	activities activities-stats attica auth baloo bookmarks \
 USE_QT=		5compat base declarative svg wayland
 USE_XORG=	x11 xcb xcursor xext xfixes xi xrender
 
-OPTIONS_DEFINE=		SDL X11
-OPTIONS_DEFAULT=	X11
+OPTIONS_DEFINE=		SDL
 OPTIONS_GROUP=		IM
 IM_DESC=		Input Method Support
 OPTIONS_GROUP_IM=	IBUS SCIM
@@ -56,7 +55,5 @@ SDL_DESC=		Enable gamecontroller System Settings module
 SDL_USES=		sdl
 SDL_USE=		sdl=sdl2
 SDL_CMAKE_BOOL_OFF=	CMAKE_DISABLE_FIND_PACKAGE_SDL2
-
-X11_RUN_DEPENDS=	xf86-input-libinput>0:x11-drivers/xf86-input-libinput
 
 .include <bsd.port.mk>


arrowd:
>  x11-drivers/xf86-input-libinput: Depends in runtime on x11-servers/xorg-server 

That's the real problem, it shouldn't. Unfortunately, fixing it requires a USES=xorg overhaul. I'll put it into my TODO, but can't say when I'll really get to it.

b-aaz:
@arrowd It would've been much appreciated if you left the PR open until further discussions conclude.
> That's the real problem, it shouldn't. Unfortunately, fixing it requires a USES=xorg overhaul. I'll put it into my TODO, but can't say when I'll really get to it.

I don't think this PR has anything to do with `xorg.mk`, as demonstrated in the patch, just removing the `RUN_DEPENDS` line on `x11-drivers/xf86-input-libinput` simply fixes the entire issue.
At one time I even hosted a binary version of [`x11/plasma6-plasma-desktop` on my overlay repo's CI](https://github.com/b-aaz/xlibre-ports/commit/87098cd8e953c4caee4c28eed3b1aeb52bc5748b) that solved this issue by just removing that single line.

_The_ problem here is that KDE plasma desktop is mistakenly & explicitly depending on an X server driver in _**runtime**_; KDE is not an X server, it can't dynamically load X server drivers in runtime, it doesn't need X server drivers in runtime, and virtually no other port/package tree in any other operating system has this runtime dependency.

Even if we were to grant that the issue is with "`xorg.mk` adding a runtime dependency on `xorg-server` to `xf86-`* drivers"  [^1] [^2] and that there is a plan to remove such runtime dependencies; this issue __would still persist__ due to `x11-drivers/xf86-input-libinput` & `x11-drivers/xlibre-xf86-input-libinput` also conflicting in runtime because they **have to** [^3] install files under similar paths.

And if we were to somehow again grant all of the above, removing an objectively superfluous runtime dependency now, with a ready made `+1 -4` patch is a much simpler, easier & faster solution than waiting for an entire "`xorg.mk` overhaul" to go through at some point in the future.
All such lengthy plans for non-related overhauls can be safely done at a later time, after fixing this active conflict which has already been waiting for **8 months+**.

In case there are any issues with the patch itself please notify me.

[^1]: Having the X server drivers depend in runtime on their X server makes sense. An `xf86-`* driver installed alone on a system is useless, they need their X server to be used for anything, the only place where this can make sense is with the `nvidia-driver`* series.
[^2]: Pedantically, [`xorg-cat.mk` is the place](https://github.com/freebsd/freebsd-ports/blob/266679a92e7b0e9c8a3d1a69bea4d504c0841e68/Mk/Uses/xorg-cat.mk#L124) that adds such dependency, _using_ `xorg.mk`.  
[^3]: Both of these drivers have to install their headers under the same name to be discovered by build systems that depend on them: [EXHIBIT A](https://github.com/freebsd/freebsd-ports/blob/266679a92e7b0e9c8a3d1a69bea4d504c0841e68/x11-drivers/xf86-input-libinput/pkg-plist#L3), [EXHIBIT B](https://github.com/freebsd/freebsd-ports/blob/266679a92e7b0e9c8a3d1a69bea4d504c0841e68/x11-drivers/xlibre-xf86-input-libinput/pkg-plist#L2)

arrowd:
> The problem here is that KDE plasma desktop is mistakenly & explicitly depending on an X server driver in runtime;

That's not true. The problem is what I wrote in my previous comment - xorg drivers should not run-depend on server.

> KDE is not an X server, it can't dynamically load X server drivers in runtime, 

This statement entails nothing.

> it doesn't need X server drivers in runtime

That's not true. The driver is required for the Touchpad KCM to work properly: https://invent.kde.org/plasma/plasma-desktop/-/blob/e66bdbe6682fa3949f6c7e227c62cc1003238163/kcms/touchpad/backends/x11/xlibbackend.cpp#L87

> this issue would still persist due to x11-drivers/xf86-input-libinput & x11-drivers/xlibre-xf86-input-libinput also conflicting in runtime because they have to [3](https://github.com/freebsd/freebsd-ports/pull/578#user-content-fn-3-c95cc99ec56d0524d69be6b0034cf082) install files under similar paths.

And this is thanks to how xlibre was implement into ports - without a proper review and communication. At the very least it should've been hooked into `bsd.default-versions.mk` so that all deps would get resolved through it.

> a ready made +1 -4 patch is a much simpler, easier & faster solution

It is not a solution, it is a hack that hurts Xorg users, which I believe are still the majority. First Xlibre is forced on us without communication and now we have to hack our ports to accommodate for poor implementation. Sorry, but no. I'm open for spending my time reviewing the framework fixes, though, unless those are LLM-generated.

PR #581, opened 2026-08-13
graphics/rawtherapee: update to 5.13

mandree:
Extend patch for rtengine/LibRaw.cmake to properly forward our local CXXFLAGS (available through ${CMAKE_CXX_FLAGS} inside cmake) to the libraw build, else it will build against GCC's default Standard C++ Library, which might be libstdc++, which then breaks when doing the final link against the common libc++ based system (GNU libstdc++ and LLVM's libc++ have incompatible ABIs).

Switch to using the system libfmt, which is version 12 and good.

Changelog:	https://rawtherapee.com/downloads/5.13/
PR:		297513


nxjoseph:
Merged, thanks.

PR #582, opened 2026-08-14
x11-servers/xorg-server: fix server stopping when started in the background

Defenso-QTH:
Starting the server in a background process group on the VT it is about to take over -- "X vtN &", or a display manager such as Ly that forks and runs the server in the child -- leaves it stopped (state T) with the console in KD_GRAPHICS, and its log ends at "(++) using VT number N".

Our existing patch calls tcsetattr() to put the tty in raw mode but ioctl TIOCSETA triggers SIGTTOU in tty_wait_background.

Do what the Linux code does instead: join the parent process group so we are not a group leader (setsid() fails for one), then setsid() into a new session with no controlling terminal.

Sponsored by: Defenso


arrowd:
While OK at first glance, it actually breaks `startx` with ConsoleKit. My use case:

```
% cat ~/.xinitrc 
exec dbus-launch --exit-with-x11 ck-launch-session xterm
% startx
===> get into graphics
% ck-list-sessions
Session1:
        unix-user = '1001'
        realname = 'arrowd'
        seat = 'seat1'
        session-type = 'unspecified'
        session-class = 'user'
        session-state = 'online'
        session-service = 'unspecified'
        active = FALSE
        x11-display = ':0'
        x11-display-device = '/dev/   ?   '
        display-device = '/dev/ttyv0'
        remote-host-name = ''
        is-local = TRUE
        on-since = '2026-08-14T17:35:23.770571Z'
        login-session-id = ''
        XDG_RUNTIME_DIR = '/var/run/xdg/arr'
```
Note `seat = 'seat1'`, `active = FALSE`, `x11-display-device = '/dev/   ?   '` - these are signs that something gone wrong.

Without this change (the "ok" case) I see this:
```
Session1:
        unix-user = '1001'
        realname = 'arrowd'
        seat = 'seat0'
        session-type = 'unspecified'
        session-class = 'user'
        session-state = 'active'
        session-service = 'unspecified'
        active = TRUE
        x11-display = ':0'
        x11-display-device = '/dev/ttyv8'
        display-device = '/dev/ttyv0'
        remote-host-name = ''
        is-local = TRUE
        on-since = '2026-08-14T17:31:38.239194Z'
        login-session-id = ''
        XDG_RUNTIME_DIR = '/var/run/xdg/arr'
        VTNr = '9'
```

I noticed an interesting "if" block in `startx`, that might be related:
```
if [ "$(uname -s)" = "Linux" ] ; then
    # When starting the defaultserver start X on the current tty to avoid
    # the startx session being seen as inactive:
    # "https://bugzilla.redhat.com/show_bug.cgi?id=806491"
    tty=$(tty)
    if expr "$tty" : '/dev/tty[0-9][0-9]*$' > /dev/null; then
        tty_num="${tty#/dev/tty}"
        vtarg="vt$tty_num -keeptty"
    fi
fi

    # For compatibility reasons, only use xserverrc if there were no server command line arguments
    if [ "$serverargs" = "" ] && [ "$display" = "" ]; then
        if [ -f "$userserverrc" ]; then
            server="$userserverrc"
        elif [ -f "$sysserverrc" ]; then
            server="$sysserverrc"
        fi
    fi
fi
```
The bugzilla link talks about a situation similar to your case. Can you please research it?

PR #595, opened 2026-08-22
misc/llama-model: new port

dennisfriedrichsen:
## Description

llama-model is a POSIX sh wrapper around llama.cpp's llama-server. It gives
local GGUF model files short, stable aliases and assembles the right
llama-server invocation - model path, mmproj path, and layered argument
files - for each alias. Each alias is a directory under a catalog root
containing a symlink to the downloaded model file (and, for a vision model,
a second symlink to its multimodal projector), so a model keeps its
descriptive download name while commands use a memorable alias. Models can
also be run detached as background services, with commands to list, start,
stop, restart, and tail their logs.

WWW: https://github.com/dennisfriedrichsen/llama-model

## Testing

- [x] `portlint -A -C -DDEVELOPER=yes`: looks fine
- [x] `make checksubdirs` in misc/: clean (SUBDIR entry correctly placed/sorted)
- [x] Full `fetch` -> `checksum` -> `extract` -> `patch` -> `stage` -> `package` pipeline succeeds with no plist warnings
- [x] Installed the built package, ran `llama-model --version`, `man llama-model`, confirmed DOCS/EXAMPLES options install correctly, then cleanly uninstalled

### DIFF
--- a/misc/Makefile
+++ b/misc/Makefile
@@ -288,6 +288,7 @@
     SUBDIR += linux-rl9-sdl12-extralibs
     SUBDIR += linux-rl9-sdl20-extralibs
     SUBDIR += llama-cpp
+    SUBDIR += llama-model
     SUBDIR += locale-en_DK
     SUBDIR += localedata
     SUBDIR += logsurfer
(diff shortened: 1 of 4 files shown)


dennisfriedrichsen:
created on accident

PR #600, opened 2026-08-27
multimedia/butt: fix build failure due to JACK

joshuamegnauth54:
BUTT depends on `audio/portaudio` with JACK enabled as of 1.47. PortAudio isn't built with JACK support by default, so BUTT's build fails in CI.

I added an option to toggle JACK which fixes the build by making the new requirement explicit. I patched the source to disable the hard dependency on JACK - it was already disabled for Windows and macOS anyway so that was easy to do.

JACK is not enabled by default in the port because audio/portaudio requires JACK enabled as well.

---

Additional context that I'm not including in the commit message.

* [Ports Fallout](https://portsfallout.com/port/multimedia/butt/)
* I tested it with poudriere with a FreeBSD 15 jail. I did not test other architectures.
* This package name is way too funny.

### DIFF
--- a/multimedia/butt/Makefile
+++ b/multimedia/butt/Makefile
@@ -1,5 +1,6 @@
 PORTNAME=	butt
 PORTVERSION=	1.47.0
+PORTREVISION=	1
 CATEGORIES=	multimedia graphics
 MASTER_SITES=	https://danielnoethen.de/butt/release/${PORTVERSION}/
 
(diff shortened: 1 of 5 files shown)


joshuamegnauth54:
@opsec Hi! Sorry for the tag and thanks for your time! After looking at other PRs, it seems like it's best practice to tag maintainers whose ports I've modified. :grin:

opsec:
This is a difficult problem: Adding a port option JACK to butt looks like it solves the problem, but due to the architecture of the ports system, it can not solve the problem that the audio/portaudio port needs to be build with the JACK option to allow the butt port to build with the JACK option. The ports system does not provide any means to allow this, as the option of one port is not coupled with (even the same) option of another port.

diizzy@ [suggested](https://lists.freebsd.org/archives/freebsd-multimedia/2026-August/004526.html) on multimedia@ to just add JACK to the default options for portaudio, but this opens another can of worms - in the end the port tree would end up with all options always (in the default tree) enabled. There were some compelling arguments against this type of 'solving this problem' in the past. Currently, there's no answer for that.

joshuamegnauth54:
Thanks for the review and sorry for the hassle! I agree with everything you said. I'll close this then work on something simpler. This is a learning process for me, and so I greatly appreciate your detailed comments. :grin:

PR #607, opened 2026-09-01
x11-wm/lxappearance-obconf: do not pollute RUN_DEPENDS

s3erios:
Currently it includes gmake, pkgconf and intltool as runtime dependencies due to incorrect variables usage

Replaced the construction with an extra OTHER_DEPENDS variable accordingly to 5.10.2. RUN_DEPENDS FreeBSD Porter's Handbook


arrowd:
Merged, thank you.

PR #615, opened 2026-09-08
misc/py-crewai-tools: use pytest-asyncio for tests

liang0417:
## Summary

- replace the AnyIO test dependency with pytest-asyncio, matching the upstream test markers
- disable plugin autoload and explicitly load pytest_asyncio.plugin
- remove the obsolete comment referencing crewAIInc/crewAI#7305

## Rationale

The two asynchronous WaitTool tests are marked with pytest.mark.asyncio. Upstream clarified in crewAIInc/crewAI#7305 and #7307 that the suite uses pytest-asyncio and that downstream test runs should load that plugin rather than switch those tests to AnyIO.

With plugin autoload disabled, the importable pytest plugin entry point is pytest_asyncio.plugin.

## Verification

- Targeted PyPI sdist command: python -m pytest --disable-plugin-autoload -p pytest_asyncio.plugin tests/tools/wait_tool_test.py -q
- Result: 23 passed
- Broader Windows run with the port ignore list: 535 passed, 2 skipped, and 6 unrelated Windows file permission/path failures

FreeBSD/poudriere verification has not been run locally.


yurivict:
I will take care of this tomorrow.

PR #616, opened 2026-09-08
devel/kcov: Update to 43

arrowd:
This fixes build using CMake 4.

poud testport'ed on FreeBSD 14 and 15.

@asomers please review.


asomers (approved):
LGTM.

However, I no longer use this port personally, because it no longer works with recent Rust toolchains.  It hasn't for about 3 years.  And it ceased to be the best way to do coverage checking a while before that.  Would you like to take over maintainership?

arrowd:
Nope. We can just return it to the pool.

asomers:
> Nope. We can just return it to the pool.

Ok, I'll do that.  Go ahead and commit your change first, and then I'll relinquish maintainership.

arrowd:
Merged, thanks.

